PR #34, opened 2022-01-19
Wierd behaviour on extensionLogin

mihairadulescu:
We tried to add this new package in maiar.art , and it has a weird behaviour, it redirects before dispatching.


src/services/login/useExtensionLogin.ts
```diff
@@ -70,11 +70,12 @@ export const useExtensionLogin = ({
         ...(signature ? { signature } : {}),
         ...(token ? { loginToken: token } : {})
       });
-
-      window.location.href = `${url.pathname}?${nextUrlParams}`;
```
mihairadulescu, line 74: dispatch is not finishing and it already redirected to the new location.
If I added a setTimeout and changed the order things worked for us.
hope it helps

mihairadulescu:
not mergeable....

StanislavSava:
This issue is fixed in 0.4.0 version, please give it a shot

PR #45, opened 2022-01-24
Set chainID from config

mihaieremia:
After a day of trying to switch the chainID from config I spotted this bug.
Not sure if is intended or I am missing something else, but if this is not a bug then please document how to change the chainID :)


StanislavSava:
This issue is fixed in 0.4.0 version, please give it a shot

PR #74, opened 2022-02-10
Realtime Transaction Callback Update

MiroMargineanu:


README.md
```diff
@@ -64,7 +68,9 @@ This library was built for applications that use React, it might not be suitable
  </summary>
 
 ### DappProvider
+
 You need to wrap your application with the **DappProvider** component, which is exported by the library, as we need to create a global Context to be able to manipulate the data.
+
```
StanislavSava, line 73: are these spaces intentional?

example/src/pages/Dashboard/Transactions/index.tsx
```diff
@@ -22,14 +29,15 @@ const Transactions = () => {
...
+      refreshAccount();
       setState({
         transactions: data,
         transactionsFetched: success
       });
     });
   };
 
-  React.useEffect(fetchData, []);
+  React.useEffect(fetchData, [transactionsStatuses.successful]);
```
StanislavSava, line 40: Please make sure that fetchData runs only if successful is true, not on false, that would be an extra request

README.md
MiroMargineanu: No, must be related to my code editor.

example/src/pages/Dashboard/Transactions/index.tsx
MiroMargineanu: I tested this but we've got an issue with the transactions not loading anymore on the initial load because the boolean return `false` when you first load the page.

PR #76, opened 2022-02-10
Realtime Transactions Callback Update

MiroMargineanu:


example/src/pages/Dashboard/Transactions/index.tsx
```diff
@@ -16,20 +23,23 @@ const Transactions = () => {
...
+          transactions: data,
+          transactionsFetched: success
+        });
       });
-    });
+    }
   };
 
-  React.useEffect(fetchData, []);
+  React.useEffect(fetchData, [successful]);
```
StanislavSava, line 42: pass in hasActivetransactions, since you're using that inside the hook

StanislavSava (changes requested):

PR #87, opened 2022-02-14
Feature/extra context

arhtudormorar:
Initial PR on external provider
 - log method is used only for testing communication


package.json
```diff
@@ -70,7 +70,7 @@
     "transaction"
   ],
   "peerDependencies": {
-    "react": "~17.0"
+    "react": "17.0.2"
```
StanislavSava, line 73: let's make it ~, so we don't depend on an absolute version

StanislavSava (changes requested):

src/UI/PageState/types.ts
```diff
@@ -1,5 +1,5 @@
 export interface PageStateProps {
-  title?: string;
+  title?: React.ReactNode;
```
StanislavSava, line 2: Maybe we should allows both string and React node?

src/wrappers/AppInitializer.tsx
```diff
@@ -1,30 +1,48 @@
...
   const [initialized, setInitialized] = useState(false);
   const account = useGetAccountInfo();
   const { address, publicKey } = account;
   const dispatch = useDispatch();
 
-  async function initializeApp() {
+  useEffect(() => {
     dispatch(initializeNetworkConfig(networkConfig));
-    setInitialized(true);
+    if (extraActions !== undefined) {
```
StanislavSava, line 28: let's use != null, it both checks for null and undefined

src/wrappers/AppInitializer.tsx
```diff
@@ -1,30 +1,48 @@
...
+      extraActions.init({
```
StanislavSava, line 29: I think it would be a good idea to use the optional chaining to avoid crashing the app. Maybe also wrap all this in a try catch to avoid crashing the whole app if this initialization fails

constantintovisi (approved):

src/UI/PageState/types.ts
arhtudormorar: String is a valid React.ReactNode, so both work

src/wrappers/AppInitializer.tsx
arhtudormorar: Done

src/wrappers/AppInitializer.tsx
arhtudormorar: OK - added optional chaining.
OnLogin is strongly typed so I guess no errors can occur there. But we can add try catch blocks if we identify possible error places

arhtudormorar: OK

package.json
arhtudormorar: Done

PR #97, opened 2022-02-17
Feature/extra context

arhtudormorar:


src/UI/SignTransactionsModals/index.tsx
```diff
@@ -66,22 +66,35 @@ function SignTransactionsModals({
     callbackRoute,
     className
   };
+
+  const LedgerConfirm = CustomConfirmScreens?.Ledger ? (
+    <CustomConfirmScreens.Ledger {...signProps} />
+  ) : (
+    <SignWithLedgerModal {...signProps} />
+  );
+  const WalletconnectConfirm = CustomConfirmScreens?.WalletConnect ? (
```
StanislavSava, line 75: I would avoid doing this outside the render, this will render the components for all providers, having it inline with && will shortcircuit it before the && condition. For example, if you sign with extension, the <SignWithWalletCOnnect> will also trigger in this case, which might have unpleasant side-effects

src/hooks/transactions/useSignTransactionsWithLedger.tsx
```diff
@@ -115,7 +118,10 @@ export function useSignTransactionsWithLedger({
             sessionId,
             status: TransactionBatchStatusesEnum.signed,
             transactions: Object.values(newSignedTransactions).map((tx) =>
-              parseTransactionAfterSigning(tx as Transaction, true)
+              parseTransactionAfterSigning(
+                tx as Transaction,
+                getIsProviderEqualTo(LoginMethodsEnum.ledger)
```
StanislavSava, line 123: we have this is a constant already, ledgerWaiting

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -66,22 +66,35 @@ function SignTransactionsModals({
...
+    <CustomConfirmScreens.WalletConnect {...signProps} />
```
StanislavSava, line 76: I would avoid doing this outside the render, this will render the components for all providers, having it inline with && will shortcircuit it before the && condition. For example, if you sign with extension, the will also trigger in this case, which might have unpleasant side-effects

StanislavSava (changes requested):

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -66,24 +66,43 @@ function SignTransactionsModals({
...
-      {getIsProviderEqualTo(LoginMethodsEnum.extension) &&
-        (CustomConfirmScreens?.Extension?.(signProps) || (
-          <SignWithExtensionModal {...signProps} />
-        ))}
-      {getIsProviderEqualTo(LoginMethodsEnum.extra) &&
-        CustomConfirmScreens?.Extra?.(signProps)}
-    </React.Fragment>
-  ) : null;
+
+  const ready = signError || hasTransactions;
```
StanislavSava, line 70: ready implies a loading of some kind, maybe "shouldRenderModals"?

StanislavSava (approved):

PR #111, opened 2022-02-25
add support for React Native usage (except for persistence)

StanislavSava:
this PR:
- fixes error on RN because of redux-persist. Now the persist is conditioned by presence of localStorage and sessionStorage on the window object;
- exposes apiCalls for reuse across apps and the fallback configs;
- allows passing in EnvironmentEnum directly;

### DIFF
--- a/src/apiCalls/transactions/sendSignedTransactions.ts
+++ b/src/apiCalls/transactions/sendSignedTransactions.ts
@@ -1,0 +1,22 @@
+import { Transaction } from '@elrondnetwork/erdjs';
+import axios from 'axios';
+import { networkSelector } from 'redux/selectors';
+import { store } from 'redux/store';
+
+export type SendSignedTransactionsReturnType = string[];
+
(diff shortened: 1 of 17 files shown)


src/apiCalls/transactions/sendSignedTransactions.ts
```diff
@@ -0,0 +1,22 @@
+import { Transaction } from '@elrondnetwork/erdjs';
+import axios from 'axios';
+import { networkSelector } from 'redux/selectors';
+import { store } from 'redux/store';
+
+export type SendSignedTransactionsReturnType = string[];
```
IagaruEustatiu, line 6: looks like SendSignedTransactionsReturnType is just an array of strings, do we need a type for it?  let's stick with build-in types, and just use Array<strings> for now, and just use types for where there is a need for it

IagaruEustatiu (changes requested):

IagaruEustatiu (approved):

PR #112, opened 2022-02-25
[*] refactor use sign txs

IagaruEustatiu:


src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -1,189 +1,236 @@
...
+      console.error(errorMessage, err);
+
     }
-  }, [transactionsToSign?.sessionId]);
-  const hasTransactions = transactionsToSign?.transactions;
+  };
+
+  useEffect(() => {
+
+    signTransactions();
```
StanislavSava, line 219: what if transactionsToSign is null? This will go into signTransactions on crash on destructuring

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -1,189 +1,236 @@
...
+    };
+    
+  };
+
+  const signTransactions = async () => {
+
+    const hasSessionId = Boolean(transactionsToSign?.sessionId);
+    const { sessionId, transactions, callbackRoute } = transactionsToSign!;
+    const hasTransactions = Boolean(transactionsToSign?.transactions?.length);
+    const isSignEligible = transactionsToSign && hasSessionId && hasTransactions;
```
StanislavSava, line 134: hasSessionId and hasTransactions will be false if this is null, the check seems redundant

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -1,189 +1,236 @@
...
+    //move to errors constants
+    const missingProviderErrorMessage = 'You need a signer/valid signer to send a transaction,use either WalletProvider, LedgerProvider or WalletConnect';
```
StanislavSava, line 136: This can probably be pulled out of the component altogether

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -1,189 +1,236 @@
...
+      const callbackUrl = `${window.location.origin}${callbackRoute}`;
+      const buildedCallbackUrl = buildReplyUrl({ callbackUrl, urlParams });
+
+      provider.signTransactions(transactions, {
+        callbackUrl: encodeURIComponent(buildedCallbackUrl)
+      });
+      
+    };
+
+    const mapTransactionsNonces = ( latestNonce: number, transactions: Array<Transaction> ): Array<Transaction> => {
```
StanislavSava, line 167: maybe rename this to something like "setTransactionNonces"

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -1,189 +1,236 @@
...
+
+   
+    /*
+    * if the transaction is cancelled 
+    * the callback will go to undefined,
+    * we save the most recent one for a valid transaction
+    */
+    savedCallback.current = callbackRoute || window.location.pathname;
+
+    const signWithWallet = ( transactions: Array<Transaction> ) => {
```
StanislavSava, line 155: nonono, don't declare functions inside functions, this can create a bunch of problems and a strain on memory. I think a better way would be to pull them out and pass the necessary arguments

StanislavSava (changes requested):

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -1,189 +1,236 @@
...
-  }
+    ExtensionProvider.getInstance()?.cancelAction?.();
+  
+  };
 
-  function onAbort(sessionId?: string) {
-    setError(null);
+  const onCancel = (errorMessage: string, sessionId?: string) => {
+
+    const isTxCancelled = errorMessage !== 'Transaction cancelled';
```
StanislavSava, line 50: Let's pull this string into a variable outside of the component

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -1,189 +1,236 @@
...
-    } catch (err) {
-      const errorMessage =
-        (error as unknown as Error)?.message ||
-        (error as string) ||
-        'error signing transaction';
-      console.error('error signing transaction', errorMessage);
+
+    } catch (error) {
+
+      const customErrorMessage = 'provider not intialized';
```
StanislavSava, line 85: maybe we should pull all these strings inside variables outside of the component, looks a lot cleaner

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -5,185 +5,228 @@ import {
...
+        return;
       }
-    } catch (err) {
+    } catch (error) {
       const errorMessage =
         (error as unknown as Error)?.message ||
         (error as string) ||
-        'error signing transaction';
-      console.error('error signing transaction', errorMessage);
+        errorsMessages.PROVIDER_NOT_INTIALIZED;
```
StanislavSava, line 114: there's a little thing that I've added on useSignWithLedger, but seems to be missing here. What do you say if we add this error to redux, so it can be read?

  dispatch(setSignTransactionsError((err as unknown as Error).message));

StanislavSava (approved):

PR #117, opened 2022-03-01
fix: window undefined for nextjs apps

schmitzt3:
This check for undefined isn't needed for TS, but JS consumers of this library have to check for undefined. Please let me know if there is anything I can do. 🍻

### DIFF
--- a/src/redux/reducers.ts
+++ b/src/redux/reducers.ts
@@ -19,7 +19,7 @@ const reducers = {
   transactionsInfo
 };
 
-if (window?.localStorage != null) {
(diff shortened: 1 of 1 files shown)


StanislavSava:
!= null checks for both null and undefined, this seems like a redundant check
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality#description

schmitzt3:
@StanislavSava sorry... super late night. You're right. I'll close this out and figure out exactly why this isn't working with nextjs on v1.0.11.

PR #125, opened 2022-03-04
[*] Logout callback wallet deauthentication fix.

MiroMargineanu:


src/utils/logout.ts
```diff
@@ -9,17 +9,19 @@ export async function logout(
...
     return;
   }
+
   store.dispatch(logoutAction());
+
   try {
     await provider.logout({ callbackUrl });
+
     if (callbackUrl) {
       if (onRedirect) {
```
IagaruEustatiu, line 23: can we merge those two ifs in one ?

IagaruEustatiu (changes requested):

src/utils/logout.ts
```diff
@@ -9,17 +9,19 @@ export async function logout(
...
         onRedirect(callbackUrl);
-      } else {
```
StanislavSava, line 21: We need to ignore this logic only for the wallet, we need it for other providers

StanislavSava (changes requested):

src/utils/logout.ts
MiroMargineanu: No longer necessary.

src/utils/logout.ts
MiroMargineanu: Done.

src/utils/logout.ts
```diff
@@ -2,23 +2,28 @@ import { logoutAction } from 'redux/commonActions';
...
+      } else if (getProviderType(provider) !== 'wallet') {
```
StanislavSava, line 26: don't use strings, we have a function getIsProviderEqualTo(LoginMethodsEnum.wallet)

StanislavSava (approved):

IagaruEustatiu (approved):

PR #128, opened 2022-03-07
[*] Transaction toast link double slash fix.

MiroMargineanu:


IagaruEustatiu (approved):

StanislavSava:
This is not actually correct, the /transactions has to stay there, because the explorer address looks like this on the server:
 explorerAddress: 'http://devnet-explorer.elrond.com',

PR #158, opened 2022-03-21
remove providers from redux logic and create separate files for each provider

StanislavSava:


src/providers/apiProvider.ts
```diff
@@ -0,0 +1,24 @@
...
+    timeout: Number(initializationNetworkConfig.apiTimeout)
+  });
+
+  return apiProvider;
+}
+
+export function getApiProvider(): ApiProvider {
+  if (apiProvider == null) {
+    return initializeApiProvider();
+  } else {
```
arhtudormorar, line 21: ternary or direct return

ioandanpopa (approved):

src/utils/account/getAccountBalance.tsx
```diff
@@ -9,5 +9,5 @@ export async function getAccountBalance(address?: string) {
     accountAddress = account.address;
   }
   const account = await getAccount(accountAddress);
-  return account.balance.toString();
+  return account?.balance?.toString() || '...';
```
arhtudormorar, line 12: maybe problem with denominate

arhtudormorar (approved):

PR #162, opened 2022-03-22
Add lifetime to TransactionToast

Theo-Farnole:


Theo-Farnole:
I fixed the package-lock.json modifcations

Theo-Farnole:
The purpose of this pull request is that the TransactionToast UI never disappear. So I added this setting to TransactionToastsList

StanislavSava (approved):

PR #192, opened 2022-04-14
Edit constants

arhtudormorar:


src/constants/network.ts
```diff
@@ -4,7 +4,10 @@ export const DEFAULT_MIN_GAS_LIMIT = 50_000;
 
 export const configEndpoint = 'dapp/config';
 
-export const fallbackNetworkConfigurations: Record<string, NetworkType> = {
+export const fallbackNetworkConfigurations: Record<
+  'mainnet' | 'testnet' | 'devnet',
```
StanislavSava, line 8: isn't this in environmentTypesEnum?

StanislavSava (approved):

src/constants/network.ts
arhtudormorar: done

PR #209, opened 2022-05-03
Idle Timer functionality

arhtudormorar:


src/wrappers/helpers/IdleTimer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const isLoggedIn = getIsLoggedIn();
+
+  const handleOnIdle = () => {
+    if (isLoggedIn) {
+      logout();
+    }
+  };
+
+  const { isIdle } = useIdleTimer({
+    timeout: 1000 * 60 * minutes,
```
StanislavSava, line 17: This could probably be pulled outside of the component into a namedVariable

src/wrappers/helpers/IdleTimer.tsx
```diff
@@ -0,0 +1,31 @@
...
+    onIdle: handleOnIdle,
+    debounce: 500
```
StanislavSava, line 19: Same here

src/wrappers/helpers/IdleTimer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  const userIsActive = !isIdle();
```
StanislavSava, line 22: This is pretty confusing, as it inverts a boolean. Maybe we should go with isUserIdle and do !(isUserIdle && isLoggedIn) below

PR #211, opened 2022-05-04
[*] Move IdleTimer to DappUI

arhtudormorar:


src/UI/IdleTimer/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  const onIdle = () => {
+    if (isLoggedIn) {
+      logout();
+    }
+  };
+
+  useIdleTimer({ timeout, onIdle, debounce });
+
+  return null;
```
StanislavSava, line 26: always returns null?

StanislavSava: It should probably be a hook then, instead of a component

src/UI/IdleTimer/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+const debounce = 500;
+
+interface IdleTimerType {
+  minutes?: number;
+  onLogout?: (props?: any) => void;
+}
+
+export const IdleTimer = ({ minutes = 10, onLogout }: IdleTimerType) => {
+  const isLoggedIn = getIsLoggedIn();
+  const defaultLogout = () => dappLogout();
```
StanislavSava, line 14: I think you can just do  const logout = onLogout ? onLogout : dappLogout and skip this part

PR #237, opened 2022-05-25
TimeRemaining helper

arhtudormorar:


src/utils/operations/timeRemaining.ts
```diff
@@ -0,0 +1,51 @@
+import moment from 'moment';
+
+export function dhms(ms: number) {
```
StanislavSava, line 3: This is a very weird name, can we expand it a bit?

src/utils/operations/timeRemaining.ts
```diff
@@ -0,0 +1,51 @@
...
+  const secs = Math.floor(minsms / 1000);
+
+  // let diff = ' ago';
+  const diff = '';
+  let secsString = secs + ' sec';
+  let minsString = mins + ' min';
+  let hrsString = hrs + ' hr';
+  let daysString = days + ' day';
+
+  if (secs > 1) secsString = secs + ' secs';
```
StanislavSava, line 19: I think we should add brackets here

PR #238, opened 2022-05-27
[*] Web wallet triggering of onCancel and onFailed transactions

arhtudormorar:
When user goes to web wallet and Cancels transactions, upon return, callbacks are no longer registered because user left the page. This PR re-enables the callbacks by restoring the previous session

### DIFF
--- a/src/services/transactions/hooks/useTrackTransactionStatus.ts
+++ b/src/services/transactions/hooks/useTrackTransactionStatus.ts
@@ -1,14 +1,20 @@
 import { useEffect } from 'react';
+import { useGetSignedTransactions } from 'hooks/transactions/useGetSignedTransactions';
 import { useSelector } from 'redux/DappProviderContext';
 import { transactionStatusSelector } from 'redux/selectors';
 import { RootState } from 'redux/store';
-import { TransactionBatchStatusesEnum } from 'types/enums';
+import { LoginMethodsEnum, TransactionBatchStatusesEnum } from 'types/enums';
 import {
+  getIsProviderEqualTo,
   getIsTransactionFailed,
   getIsTransactionPending,
   getIsTransactionSuccessful,
   getIsTransactionTimedOut
 } from 'utils';
+import {
+  removeAllSignedTransactions,
+  removeAllTransactionsToSign
+} from '../clearTransactions';
 
 export interface UseTrackTransactionStatusArgsType {
   transactionId: string | null;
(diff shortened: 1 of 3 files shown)


src/services/transactions/hooks/useTrackTransactionStatus.ts
```diff
@@ -19,12 +25,27 @@ export interface UseTrackTransactionStatusArgsType {
...
+  transactionId: txId,
   onSuccess,
   onFail,
   onCancelled,
   onTimedOut
 }: UseTrackTransactionStatusArgsType) {
+  const { signedTransactionsArray } = useGetSignedTransactions();
+  const isWalletProvider = getIsProviderEqualTo(LoginMethodsEnum.wallet);
+
+  const [sessionId] =
```
StanislavSava, line 37: I think this should be called, for clarify, "lastSessionId"

src/services/transactions/hooks/useTrackTransactionStatus.ts
```diff
@@ -19,12 +25,27 @@ export interface UseTrackTransactionStatusArgsType {
...
+    signedTransactionsArray.length > 0
+      ? signedTransactionsArray[signedTransactionsArray.length - 1]
+      : [];
+
+  /**
+   * Web wallet restores sessionId
```
StanislavSava, line 43: Can we expand the comment a bit, please? Maybe something like "when using web wallet, txId will be lost when redirected back, so we extract it from signedTransactions"

src/services/transactions/hooks/useTrackTransactionStatus.ts
```diff
@@ -19,12 +25,27 @@ export interface UseTrackTransactionStatusArgsType {
...
+   */
+  const walletSessionId = txId ?? sessionId ?? null;
+
+  const transactionId = isWalletProvider ? walletSessionId : txId;
```
StanislavSava, line 47: I think we should skip this check, if txId is null, which will be the case with wallet, it will fallback to sessionId. If not, as it is for other providers, it will not trigger the ??. 

This can even be helpful, as someone who sends a tx and forgets to track the tx will get info about the tx anyway

src/services/transactions/hooks/useTrackTransactionStatus.ts
```diff
@@ -38,6 +59,13 @@ export function useTrackTransactionStatus({
 
   const isCancelled = status === TransactionBatchStatusesEnum.cancelled;
 
+  function preventWalletDoubleCallback() {
+    if (isWalletProvider) {
```
StanislavSava, line 63: Can you drop a comment as to what this does, please?

PR #267, opened 2022-06-20
replace dts with esbuild for faster and lighter builds and replace single place of entry with multiple entries

StanislavSava:
This PR aims to solve a common problem when using dapp-core. Given that there is a single entry point for the build, all the code is bundled together and executed at runtime, regardless of what you use from there. This becomes problematic not only because of final bundle size, but also you need a lot of polyfills for non-browser environments.

This PR replaces the dts-cli build with esbuild and multiple entry/output "modules". This way you don't end up importing code you're not using.

Further down the road we aim to move all "web-specific" code to the web folder and make dapp-core as platform agnostic as possible, to allow anyone to use it without adding a bunch of shims and polyfills


esbuild.js
```diff
@@ -0,0 +1,48 @@
+const svgrPlugin = require('esbuild-plugin-svgr');
+const esbuild = require('esbuild');
+const plugin = require('node-stdlib-browser/helpers/esbuild/plugin');
+const stdLibBrowser = require('node-stdlib-browser');
+const { nodeExternalsPlugin } = require('esbuild-node-externals');
+const fs = require('fs');
+
+const entryPoints = fs
```
grigorebudac, line 8: maybe it would be a better approach to use something like [glob](https://www.npmjs.com/package/glob) for implementing the pattern matching?

package.json
```diff
@@ -6,46 +6,33 @@
   "license": "GPL-3.0-or-later",
   "repository": "ElrondNetwork/dapp-core",
   "main": "dist/index.js",
-  "module": "dist/dapp-core.esm.js",
-  "typings": "dist/index.d.ts",
+  "typings": "dist/hooks/index.d.ts",
```
grigorebudac, line 9: let's make sure the typings path is correct (it probably needs some adjustments)

package.json
```diff
@@ -117,6 +105,12 @@
       "limit": "10 KB"
     }
   ],
+  "files": [
```
grigorebudac, line 108: I'm curious if we really need the files property (will have to double-check on this)

src/UI/ProgressSteps/index.tsx
```diff
@@ -7,7 +7,7 @@ export interface ProgressStepsType {
   className?: string;
 }
 
-const ProgressSteps = ({
+export const ProgressSteps = ({
```
grigorebudac, line 10: I guess we can delete the export default then

src/UI/ExplorerLink/index.tsx
```diff
@@ -1,10 +1,10 @@
...
-import { useGetNetworkConfig } from 'hooks';
 
-import icons from 'optionalPackages/fortawesome-free-solid-svg-icons';
-import ReactFontawesome from 'optionalPackages/react-fontawesome';
+import { faSearch } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { useGetNetworkConfig } from 'hooks';
 
-const ExplorerLink = ({
+export const ExplorerLink = ({
```
grigorebudac, line 7: sam here, we can probably get rid of the export default

src/UI/SignTransactionsModals/SignWithDeviceModal/index.tsx
```diff
@@ -1,11 +1,11 @@
...
+import { Modal } from 'react-bootstrap';
 import { useSignTransactionsWithDevice } from 'hooks';
-import ReactBootstrap from 'optionalPackages/react-bootstrap';
 import { SignModalPropsType } from 'types';
-import { getGeneratedClasses } from 'utils';
+import { getGeneratedClasses } from 'UI/utils';
 import SignStep from './SignStep';
 
-const SignWithDeviceModal = ({
+export const SignWithDeviceModal = ({
```
grigorebudac, line 8: and here

src/UI/SignTransactionsModals/SignWithExtensionModal/index.tsx
```diff
@@ -1,11 +1,12 @@
...
+import { faHourglass, faTimes } from '@fortawesome/free-solid-svg-icons';
+import { Modal } from 'react-bootstrap';
 import { SignModalPropsType } from 'types';
 import PageState from 'UI/PageState';
-import { getGeneratedClasses, safeRedirect } from 'utils';
+import { getGeneratedClasses } from 'UI/utils';
+import { safeRedirect } from 'utils';
 
-const SignWithExtensionModal = ({
+export const SignWithExtensionModal = ({
```
grigorebudac, line 9: and here

src/UI/SignTransactionsModals/SignWithLedgerModal/index.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { SignModalPropsType } from 'types';
 import SignWithDeviceModal from '../SignWithDeviceModal';
 
-const SignWithLedgerModal = (props: SignModalPropsType) => {
+export const SignWithLedgerModal = (props: SignModalPropsType) => {
```
grigorebudac, line 5: and here

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -36,11 +36,11 @@ interface SignTransactionsPropsType {
   verifyReceiverScam?: SignPropsType['verifyReceiverScam'];
 }
 
-function SignTransactionsModals({
+export const SignTransactionsModals = ({
```
grigorebudac, line 39: and here

src/UI/TransactionToast/index.tsx
```diff
@@ -1,31 +1,49 @@
...
+  processingMessage?: string;
+  startTimeProgress?: number;
+  shouldRenderDefaultCss?: boolean;
+  transactions: SignedTransactionType[];
+  status: TransactionBatchStatusesEnum;
+  lifetimeAfterSuccess?: number;
+  onClose?: (toastId: string) => void;
+}
+
+export const TransactionToast = ({
```
grigorebudac, line 46: and here

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -2,18 +2,28 @@ import { useEffect, useRef, useState } from 'react';
...
 
 import { ExtensionProvider } from '@elrondnetwork/erdjs-extension-provider';
-import { errorsMessages, walletSignSession } from 'constants/index';
+import {
+  ERROR_SIGNING,
+  ERROR_SIGNING_TX,
+  MISSING_PROVIDER_MESSAGE,
+  PROVIDER_NOT_INTIALIZED,
+  TRANSACTION_CANCELLED,
+  walletSignSession
```
arhtudormorar, line 11: WALLET_SIGN_SESSION

src/index.ts
```diff
@@ -0,0 +1 @@
+export default {};
```
arhtudormorar, line 1: is this used?

src/reduxStore/middlewares/loginSessionMiddleware.ts
```diff
@@ -0,0 +1,53 @@
+import { PayloadAction } from '@reduxjs/toolkit';
+import throttle from 'lodash.throttle';
+import { getAccountProvider } from 'providers/accountProvider';
+import { logoutAction } from 'reduxStore/commonActions';
+import { isLoggedInSelector } from 'reduxStore/selectors/loginInfoSelectors';
+import { getNewLoginExpiresTimestamp, setLoginExpiresAt } from 'storage/local';
+import storage from 'utils/storage';
+import { localStorageKeys } from 'utils/storage/local';
+
+const whitelistedActions = ['logout'];
```
arhtudormorar, line 10: this string should be hardcoded in constants and referenced in slice

src/reduxStore/selectors/accountInfoSelectors.ts
```diff
@@ -1,7 +1,6 @@
-import { RootState } from '../store';
 import { createDeepEqualSelector } from './helpers';
 
-export const accountInfoSelector = (state: RootState) => state.account;
+export const accountInfoSelector = (state: any) => state.account;
```
arhtudormorar, line 3: TBD maybe find a solution not to use `any` here

src/reduxStore/selectors/loginInfoSelectors.ts
```diff
@@ -1,9 +1,8 @@
 import { LoginMethodsEnum } from 'types/enums';
-import { RootState } from '../store';
 import { addressSelector } from './accountInfoSelectors';
 import { createDeepEqualSelector } from './helpers';
 
-export const loginInfoSelector = (state: RootState) => state.loginInfo;
+export const loginInfoSelector = (state: any) => state.loginInfo;
```
arhtudormorar, line 5: TBD

src/reduxStore/store.ts
```diff
@@ -29,22 +27,18 @@ const migrations = {
...
-    whitelist: ['account', 'loginInfo', 'toasts', 'modals', 'networkConfig'],
-    migrate: createMigrate(migrations, { debug: false })
-  };
-  localStorageReducers = persistReducer(persistConfig, rootReducer);
-}
+const persistConfig: any = {
+  key: 'dapp-core-store',
+  version: 2,
+  storage,
+  whitelist: ['account', 'loginInfo', 'toasts', 'modals', 'networkConfig'],
```
arhtudormorar, line 34: Strings could be defined under constants/whitelisdedSlices and imported from there, to avoid typos

src/services/transactions/transformAndSignTransactions.ts
```diff
@@ -68,7 +66,7 @@ export async function transformAndSignTransactions({
       sender: new Address(address).hex(),
       chainID: transactionsChainId,
       version: version ?? 1,
```
arhtudormorar, line 68: can use version = 1 at line 43

package.json
```diff
@@ -6,46 +6,33 @@
...
+    "@types/qrcode": "^1.4.2",
     "@types/qs": "^6.9.7",
-    "@types/react": "18.0.10",
-    "@types/react-dom": "18.0.5",
     "@typescript-eslint/eslint-plugin": "5.0.0",
     "@typescript-eslint/parser": "5.14.0",
-    "babel-loader": "8.2.4",
-    "babel-plugin-module-resolver": "^4.1.0",
-    "dts-cli": "^1.5.1",
+    "esbuild": "^0.14.43",
```
cipriandraghici93, line 33: Maybe is better to use a strict package version. Please take into consideration also the other new packages added.

src/UI/SignTransactionsModals/SignWithDeviceModal/index.tsx
```diff
@@ -62,7 +62,7 @@ const SignWithDeviceModal = ({
           />
         </div>
       </div>
-    </ReactBootstrap.Modal>
+    </Modal>
   );
 };
```
cipriandraghici93, line 68: Still need the default export?

src/UI/SignTransactionsModals/SignWithExtensionModal/index.tsx
```diff
@@ -62,7 +63,7 @@ const SignWithExtensionModal = ({
           </button>
         }
       />
-    </ReactBootstrap.Modal>
+    </Modal>
   );
 };
```
cipriandraghici93, line 69: Still need the default export  ?

src/UI/SignTransactionsModals/SignWithWalletConnectModal/index.tsx
```diff
@@ -63,7 +64,7 @@ export const SignWithWalletConnectModal = ({
           </button>
         }
       />
-    </ReactBootstrap.Modal>
+    </Modal>
   );
 };
```
cipriandraghici93, line 70: Still need the default export ?

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -105,6 +105,6 @@ function SignTransactionsModals({
   }
 
   return null;
-}
+};
 
 export default SignTransactionsModals;
```
cipriandraghici93, line 110: I see that a lot of components still keep the default export. This is still needed ?

package.json
StanislavSava: As per package.json docs: `The optional files field is an array of file patterns that describes the entries to be included when your package is installed as a dependency. File patterns follow a similar syntax to .gitignore, but reversed: including a file, directory, or glob pattern (*, **/*, and such) will make it so that file is included in the tarball when it's packed. Omitting the field will make it default to ["*"], which means it will include all files.`

I know we have a npmignore for this, but I think this a useful property

src/index.ts
cipriandraghici93, line 1: Why you need this empty default export ?

src/reduxStore/selectors/accountInfoSelectors.ts
cipriandraghici93, line 3: `any` type should be avoided as possible. I see that before it was used `RootState`. Is this deprecated?

src/reduxStore/selectors/loginInfoSelectors.ts
cipriandraghici93, line 5: same here. Please see my comment related to any type.

cipriandraghici93 (changes requested):

src/reduxStore/selectors/accountInfoSelectors.ts
StanislavSava: adding RootState here creates a circular dependency

cipriandraghici93 (approved):

arhtudormorar (approved):

PR #272, opened 2022-06-22
move accountProvider to redux, drop proxy and Api provider

StanislavSava:


src/apiCalls/accounts/getAccountFromApi.ts
```diff
@@ -0,0 +1,29 @@
...
+import { store } from 'reduxStore/store';
+import { addressEndpoint } from 'apiCalls/endpoints';
+import { AccountType } from 'types';
+
+export async function getAccountFromApi(address?: string) {
+  if (!address) {
+    return null;
+  }
+  const apiAddress = apiAddressSelector(store.getState());
+  const cleanApiAddress = apiAddress.endsWith('/')
```
grigorebudac, line 12: it seems like this part is being copy-pasted in multiple files. I would create a new selector here, something like cleanApiAddressSelector

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -149,7 +159,7 @@ export const useSignTransactions = () => {
       }
     } catch (error) {
       const errorMessage =
-        (error as unknown as Error)?.message ||
+        ((error as unknown) as Error)?.message ||
```
grigorebudac, line 162: have you tried casting it directly as Error?

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -217,7 +216,7 @@ export const useSignTransactions = () => {
         signTransactionsWithProvider();
       }
     } catch (err) {
-      const defaultErrorMessage = (error as unknown as Error)?.message;
+      const defaultErrorMessage = ((error as unknown) as Error)?.message;
```
grigorebudac, line 219: same here

src/reduxStore/store.ts
```diff
@@ -51,9 +51,14 @@ export const store = configureStore({
           PURGE,
           REGISTER,
           'accountInfoSlice/setAccount',
-          'accountInfoSlice/setAccountNonce'
+          'accountInfoSlice/setAccountNonce',
```
grigorebudac, line 54: this approach is ok, alternatively, you can get the name dynamically from the slice

arhtudormorar, line 54: would be nice to use these strings from a config

src/types/network.ts
```diff
@@ -0,0 +1,61 @@
...
+  walletConnectDeepLink?: string;
+  walletConnectBridgeAddresses?: string[];
+  walletAddress?: string;
+  apiAddress?: string;
+  explorerAddress?: string;
+  skipFetchFromServer?: boolean;
+  apiTimeout?: string;
+}
+
+export interface BlockchainNetworkConfigType {
```
arhtudormorar, line 41: rename to ApiNetworkConfigType

src/wrappers/DappProvider/index.tsx
```diff
@@ -21,8 +17,6 @@ export interface DappProviderPropsType {
   customNetworkConfig?: CustomNetworkType;
   externalProvider?: IDappProvider;
   environment: 'testnet' | 'mainnet' | 'devnet' | EnvironmentsEnum;
```
arhtudormorar, line 19: can use `keyof typeof EnvironmentsEnum`

cipriandraghici93 (approved):

PR #274, opened 2022-06-22
move accountProvider out of redux

StanislavSava:


src/hooks/useDebounce.ts
```diff
@@ -1,6 +1,6 @@
 import { useState, useEffect } from 'react';
 
-const useDebounce = (value: any, timeout: number) => {
+export const useDebounce = (value: any, timeout: number) => {
```
arhtudormorar, line 3: would be nicer with <T>(value: T, timeout: number)

src/reduxStore/store.ts
```diff
@@ -52,8 +50,7 @@ export const store = configureStore({
           PURGE,
           REGISTER,
           setAccountNonce.type,
-          setAccount.type,
-          setAccountProvider.type
+          setAccount.type
```
arhtudormorar, line 53: ❤️

PR #283, opened 2022-06-24
remove non pending transactions on page reload or tab close

cipriandraghici93:


src/UI/TransactionsToastList/index.tsx
```diff
@@ -103,6 +119,39 @@ export const TransactionsToastList = ({
...
+      const { status } = currentTx;
+      const isPending = getIsTransactionPending(status);
+
+      if (!isPending) {
+        handleDeleteTransactionToast(transactionToast.toastId);
+      }
+    });
+  };
+
+  console.log('Outside', signedTransactionsToRenderRef.current);
```
grigorebudac, line 139: oops

src/UI/Trim/styles.scss
```diff
@@ -67,6 +67,22 @@
...
+  @media (max-width: 991.98px) {
+    max-width: 12rem !important;
+  }
+
+  @media (max-width: 768px) {
+    max-width: 8rem !important;
+  }
+}
+
+.test-class {
```
grigorebudac, line 84: I don't think you need this class anymore

src/UI/Trim/styles.scss
```diff
@@ -67,6 +67,22 @@
     display: block;
     color: transparent;
   }
+
+  @media (max-width: 1199.98px) {
+    max-width: 26rem !important;
```
grigorebudac, line 72: does it work without `!important`?

src/assets/sass/themes/table.scss
```diff
@@ -140,3 +140,29 @@
...
 }
+
+
+@media (max-width: 767.98px) {
+  .transactions.table .transaction-hash .trim {
+    max-width: 8rem;
+  }
+}
+
+//@media (max-width: 991.98px) {
```
grigorebudac, line 151: I think that you can remove this part as it's not used anymore

src/UI/Progress/index.tsx
```diff
@@ -24,104 +26,136 @@ export const Progress = ({
   done,
   expiresIn = 10 * 60
 }: ProgressProps) => {
-  const ref = React.useRef(null);
-  const intervalRef = React.useRef<any>();
-  const removeTxFromSession = () => {
-    const toastProgress = storage.session.getItem('toastProgress');
+  const { totalSeconds, currentRemaining } = getInitialData();
```
StanislavSava, line 29: I think this could be a useMemo

src/UI/TransactionsToastList/index.tsx
```diff
@@ -7,16 +7,17 @@ import {
...
 import CustomToast from 'UI/TransactionsToastList/components/CustomToast';
 import TransactionToast from 'UI/TransactionsToastList/components/TransactionToast';
-import { deleteCustomToast } from 'utils';
+import { deleteCustomToast, getIsTransactionPending } from 'utils';
 
 import styles from './styles.scss';
 import { CustomToastType, TransactionToastType } from 'types/toasts';
 import { addTransactionToast, removeTransactionToast } from 'reduxStore/slices';
 import { removeSignedTransaction } from 'services';
+import { store } from '../../reduxStore/store';
```
StanislavSava, line 20: let's use absolute imports

src/UI/TransactionsToastList/index.tsx
```diff
@@ -103,6 +119,39 @@ export const TransactionsToastList = ({
...
 
+  const clearNotPendingTransactionsFromStorage = (e: BeforeUnloadEvent) => {
+    e.preventDefault();
+
+    const toasts = transactionToastsSelector(store.getState());
+    toasts.forEach((transactionToast: TransactionToastType) => {
+      const currentTx: SignedTransactionsBodyType =
+        signedTransactionsToRenderRef.current[transactionToast.toastId];
+
+      const { status } = currentTx;
```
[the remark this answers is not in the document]
StanislavSava: We should probably handle for null here

src/UI/TransactionsToastList/index.tsx
```diff
@@ -103,6 +119,39 @@ export const TransactionsToastList = ({
     )
   );
 
+  const clearNotPendingTransactionsFromStorage = (e: BeforeUnloadEvent) => {
+    e.preventDefault();
+
+    const toasts = transactionToastsSelector(store.getState());
```
StanislavSava, line 125: I think we should move this logic to the TransactionToast component, since we have all the data there and we don't need to re-read everything on umount

src/UI/TransactionsToastList/index.tsx
StanislavSava, line 139: forgot a console.log

src/UI/Trim/styles.scss
cipriandraghici93: yes

src/UI/TransactionsToastList/index.tsx
cipriandraghici93: I do not agree with this. I want to keep a Single Responsibility for the Toast component.

StanislavSava (approved):

PR #292, opened 2022-06-29
Refactoring: extract current transaction toast logic to the parent level

cipriandraghici93:


cipriandraghici93:
Closing. A new implementation is comming.

PR #293, opened 2022-06-29
Refactoring - move current transaction logic to the parent level + fix TransactionToast wrapper style

cipriandraghici93:


src/UI/TransactionsToastList/index.tsx
```diff
@@ -83,19 +83,40 @@ export const TransactionsToastList = ({
...
-              type,
-              startTimestamp,
-              toastId,
-              signedTransactionsToRender,
-              lifetimeAfterSuccess: successfulToastLifetime,
-              onDelete: handleDeleteTransactionToast
-            }}
-          />
-        )
+        ({ toastId, type, startTimestamp }: TransactionToastType) => {
```
StanislavSava, line 86: I think it would be a good idea to pull this out into a separate function outside of the component, renderTransactionToast

StanislavSava (approved):

PR #299, opened 2022-07-04
fix login with wallet connect and all the other providers

cipriandraghici93:


src/hooks/login/useExtensionLogin.ts
```diff
@@ -23,13 +22,14 @@ export type UseExtensionLoginReturnType = [
...
   callbackRoute,
   token,
-  redirectAfterLogin = false
 }: UseExtensionLoginPropsType): UseExtensionLoginReturnType => {
   const [error, setError] = useState('');
   const [isLoading, setIsLoading] = useState(false);
   const isLoggedIn = useSelector(isLoggedInSelector);
   const dispatch = useDispatch();
 
+  const redirectAfterLogin = Boolean(callbackRoute);
```
StanislavSava, line 31: Maybe we should just move this logic inside optionalRedirect, to make sure there is no confusion

src/hooks/login/useExtensionLogin.ts
```diff
@@ -45,7 +45,7 @@ export const useExtensionLogin = ({
       }
 
       const callbackUrl: string = encodeURIComponent(
-        `${window.location.origin}${callbackRoute}`
+        `${window.location.origin}${callbackRoute ?? window.location.href}`
```
StanislavSava, line 48: we need to get the pathname here, because there's a window.location.origin prefix and this will break the URL

src/hooks/login/useLedgerLogin.ts
```diff
@@ -59,8 +58,8 @@ export function useLedgerLogin({
   callbackRoute,
   token,
   addressesPerPage = defaultAddressesPerPage,
-  redirectAfterLogin = false
 }: UseLedgerLoginPropsType): LedgerLoginHookReturnType {
+  const redirectAfterLogin = Boolean(callbackRoute);
```
StanislavSava, line 62: do we use this for anything here? If this is passed down, let's just pass callbackRoute until it's actually used

src/utils/internal/optionalRedirect.ts
```diff
@@ -1,9 +1,10 @@
 import { safeRedirect } from '../redirect';
 
 export function optionalRedirect(
-  callbackUrl?: string,
-  shouldRedirect?: boolean
+  callbackUrl?: string
 ) {
+  const shouldRedirect = Boolean(callbackUrl);
```
StanislavSava, line 6: I think we could drop this guy and move the boolean(callbackUrl) in the if below

cipriandraghici93: If I do this, the below code complains about missing callbackUrl and I have to pass the default value in 2 places. That's why I kept the validation for callbackUrl in if statement

image

cipriandraghici93: I will come with a small change here which is more elegant approach

cipriandraghici93: @StanislavSava,  what about this?

```
export function optionalRedirect(
  callbackUrl?: string
) {
  if (Boolean(callbackUrl)) {
    const callbackRoute = callbackUrl ?? window.location.href;

    setTimeout(() => {
      if (!window.location.pathname.includes(callbackRoute)) {
        safeRedirect(callbackRoute);
      }
    }, 200);
  }
}
```

PR #304, opened 2022-07-06
Feature - implement cancel and error transaction toasts

CiprianDraghici:
Cancel transaction toast:

image

Error transaction toast:

image


src/UI/SignTransactionsModals/index.tsx
```diff
@@ -11,6 +10,7 @@ import { LoginMethodsEnum } from 'types';
 import { SignWithExtensionModal } from './SignWithExtensionModal';
 import { SignWithLedgerModal } from './SignWithLedgerModal';
 import { SignWithWalletConnectModal } from './SignWithWalletConnectModal';
+import { TransactionStatusToast } from '../../components/TransactionStatusToast/TransactionStatusToast';
```
StanislavSava, line 13: let's use absolute import

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -71,32 +72,71 @@ export const SignTransactionsModals = ({
...
         ) : (
-          <SignWithExtensionModal {...signProps} />
+          <ConfirmScreens.Extension {...signProps} />
         );
 
       case LoginMethodsEnum.extra:
-        return CustomConfirmScreens?.Extra ? (
-          <CustomConfirmScreens.Extra {...signProps} />
+        return ConfirmScreens?.Extra ? (
+          shouldShowTransactionStatusToast ? (
```
StanislavSava, line 135: I think I saw a comment left by you somewhere that double ternary are confusing :)

src/UI/TransactionsToastList/index.tsx
```diff
@@ -141,8 +142,12 @@ export const TransactionsToastList = ({
     ]
   );
 
-  const customToastsList = customToasts.map(
-    ({ toastId, type, duration, message = '' }: CustomToastType) => (
+  const customToastsList = customToasts
+    .filter(
+      (customToast: CustomToastType) =>
+        customToast.toastId !== TRANSACTION_STATUS_TOAST_ID
```
StanislavSava, line 148: What does this filter do? Can we drop a commentary?

src/components/TransactionStatusToast/TransactionStatusToast.tsx
```diff
@@ -0,0 +1,130 @@
+import { addNewCustomToast, deleteCustomToast } from '../../utils';
+import { CustomToast } from '../../UI';
+import React, { useEffect, useMemo } from 'react';
+import { CustomToastType } from '../../types/toasts';
```
StanislavSava, line 4: let's replace all with absolute imports

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -71,32 +72,71 @@ export const SignTransactionsModals = ({
...
-  if (signError || hasTransactions) {
+  const ConfirmScreens = {
+    Ledger: CustomConfirmScreens?.Ledger ?? SignWithLedgerModal,
+    WalletConnect:
+      CustomConfirmScreens?.WalletConnect ?? SignWithWalletConnectModal,
+    Extension: CustomConfirmScreens?.Extension ?? SignWithExtensionModal,
+    Extra: CustomConfirmScreens?.Extra ?? null
+  };
+
+  const renderTransactionStatusToast = () => {
```
StanislavSava, line 83: I think we can turn this login into a single <TransactionStatusToast which gets different props, depending on the type of the flow

src/components/TransactionStatusToast/TransactionStatusToast.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+type ErrorToastProps = {
+  type?: 'info' | 'error' | 'warning';
+  message: string;
+  duration?: number;
+  show?: boolean;
+  onDelete?: () => void;
+};
+
+export const TransactionStatusToast: React.FC<ErrorToastProps> = ({
```
StanislavSava, line 22: This name is a bit deceiving, all toasts are transaction status toasts

src/components/TransactionStatusToast/TransactionStatusToast.tsx
```diff
@@ -0,0 +1,130 @@
...
+  show,
```
StanislavSava, line 23: I think we should invert the control and not pass a show prop here

src/components/TransactionStatusToast/TransactionStatusToast.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const errorToast = useMemo(
+    () =>
+      customToasts.find(
+        (customToast: CustomToastType) =>
+          customToast.toastId === TRANSACTION_STATUS_TOAST_ID
+      ),
+    [customToasts]
+  );
+
+  const showErrorToast = () => {
```
StanislavSava, line 40: We have the JSX below, which renders something, and we have a dispatch here that adds something to store. Does this not trigger a rendering in the TransactionToastsList or somewhere where there's a listener for this value? I think we should either dispatch an action or render something, not both

src/components/TransactionStatusToast/TransactionStatusToast.tsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+  };
+
+  useEffect(() => {
+    if (show) {
+      showErrorToast();
+    }
+  }, [show]);
+
+  const StatusIcon = useMemo(() => {
```
StanislavSava, line 63: What if we lift this up and use a renderProp or pass the Icon as a child to the Toast, to avoid rerendering this when something changes and avoid using useMemo? We can create a component called StatusIcon that receives a type and renders all this

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -50,10 +52,19 @@ export const useSignTransactions = () => {
   const { provider } = useGetAccountProvider();
   const providerType = getProviderType(provider);
   const [error, setError] = useState<string | null>(null);
+  const [cancelTransactionsMessage, setCancelTransactionsMessage] = useState<
```
StanislavSava, line 55: I think the preffix should be "cancelled"

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -162,13 +178,18 @@ export const useSignTransactions = () => {
...
+
       dispatch(
         moveTransactionsToSignedState({
           sessionId,
           status: TransactionBatchStatusesEnum.cancelled
         })
       );
-      onCancel(errorMessage, sessionId);
+      onCancel(
+        errorMessage.includes('cancel') ? TRANSACTION_CANCELLED : errorMessage,
```
StanislavSava, line 190: This string should probably be in an enum

src/UI/SignTransactionsModals/index.tsx
cipriandraghici93: right

src/UI/TransactionsToastList/index.tsx
cipriandraghici93: We are adding a new CustomToast with that specific id and we use this toast for displaying the canceled/error messages on signing a transaction. The toast is persisted in custom toasts list and we do not want to display multiple times the toast (if the TransactionToastList component is used)

cipriandraghici93: I will add a comment

src/hooks/transactions/useSignTransactions.tsx
cipriandraghici93: Unfortunately, erdjs throws inconsistent errors from different providers...it's hard to sync this logic with all errors thrown

src/components/TransactionStatusToast/types.ts
```diff
@@ -0,0 +1 @@
+export type StatusIconType = 'info' | 'error' | 'warning';
```
grigorebudac, line 1: I think that an Enum would be more suitable because you can then do something like `MyEnum.INFO` when creating switch case declarations

src/components/TransactionStatusToast/TransactionStatusToast.tsx
```diff
@@ -0,0 +1,67 @@
...
+        (customToast: CustomToastType) =>
+          customToast.toastId === TRANSACTION_STATUS_TOAST_ID
+      ),
+    [customToasts]
+  );
+
+  const handleDelete = () => {
+    if (Boolean(errorToast)) {
+      deleteCustomToast(TRANSACTION_STATUS_TOAST_ID);
+      onDelete?.();
```
grigorebudac, line 38: looks like you're running deleteCustom anyway, which means that only `onDelete` could be wrapped inside an if condition

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -123,7 +138,9 @@ export const useSignTransactions = () => {
       const errorMessage =
         (error as Error)?.message ||
         (error as string) ||
-        PROVIDER_NOT_INTIALIZED;
+        PROVIDER_NOT_INITIALIZED;
+      console.error(errorMessage);
```
grigorebudac, line 142: oops

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -162,13 +179,18 @@ export const useSignTransactions = () => {
     } catch (error) {
       const errorMessage =
         (error as Error)?.message || (error as string) || ERROR_SIGNING_TX;
+      console.error(errorMessage);
```
grigorebudac, line 182: oops

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -162,13 +179,18 @@ export const useSignTransactions = () => {
...
+
       dispatch(
         moveTransactionsToSignedState({
           sessionId,
           status: TransactionBatchStatusesEnum.cancelled
         })
       );
-      onCancel(errorMessage, sessionId);
+      onCancel(
+        errorMessage.includes('cancel') ? TRANSACTION_CANCELLED : errorMessage,
```
grigorebudac, line 191: soo, if the error message changes, this won't work?

src/hooks/transactions/useSignTransactions.tsx
```diff
@@ -218,6 +242,8 @@ export const useSignTransactions = () => {
     } catch (err) {
       const defaultErrorMessage = (err as Error)?.message;
       const errorMessage = defaultErrorMessage || ERROR_SIGNING;
+      console.error(errorMessage);
```
grigorebudac, line 245: forgot this

src/hooks/transactions/useSignTransactions.tsx
CiprianDraghici: the console error is added intentionally to log the error from the catch block

src/hooks/transactions/useSignTransactions.tsx
CiprianDraghici: Right. I don't have control over what erdjs packages throw. I tried a general solution using what is currently thrown at this moment.

src/hooks/transactions/useSignTransactions.tsx
CiprianDraghici: No, it is intentionally used.

src/UI/SignTransactionsModals/components/TxStatusToast.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const TxStatusToast: React.FC<TxStatusToastProps> = ({
+  signError,
+  canceledTransactionsMessage,
+  onDelete
+}) => {
+  if (signError) {
+    return (
+      <TransactionStatusToast
+        message={signError}
```
StanislavSava, line 18: We can render just one instance of TransactionStatusToast and just change the props

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -71,39 +58,55 @@ export const SignTransactionsModals = ({
...
-        );
+  const shouldShowTransactionStatusToast =
+    Boolean(signError) || Boolean(cancelTransactionsMessage);
 
-      case LoginMethodsEnum.extra:
-        return CustomConfirmScreens?.Extra ? (
-          <CustomConfirmScreens.Extra {...signProps} />
-        ) : null;
+  const renderScreen = useCallback(
+    (Screen?: ScreenType) => {
```
StanislavSava, line 73: This logic seems a bit crowded, a lot of features into a couple of components with a lot of props. Is there a way we can simplify this rendering, so it's easier to read?

src/components/TransactionStatusToast/TransactionStatusToast.tsx
CiprianDraghici: good catch

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -71,39 +58,55 @@ export const SignTransactionsModals = ({
...
+      return (
+        <ConfirmationScreen
+          Screen={Screen}
+          signProps={signProps}
```
StanislavSava, line 77: this function receives the exact same props, just for the 1st, which is the screen. Maybe we can make this into a JSX that receives the ConfirmScreen and tries to extras as much as possible from Redux, instead of getting prop drilled?

src/UI/TransactionsToastList/index.tsx
```diff
@@ -16,6 +16,7 @@ import { CustomToastType, TransactionToastType } from 'types/toasts';
 import { addTransactionToast, removeTransactionToast } from 'reduxStore/slices';
 import { removeSignedTransaction } from 'services';
 import { store } from 'reduxStore/store';
+import { TRANSACTION_STATUS_TOAST_ID } from '../../constants';
```
StanislavSava, line 19: absolutes. I think you should configure your IDE to default to absolute imports

src/UI/TransactionsToastList/index.tsx
```diff
@@ -141,8 +142,14 @@ export const TransactionsToastList = ({
     ]
   );
 
-  const customToastsList = customToasts.map(
-    ({ toastId, type, duration, message = '' }: CustomToastType) => (
+  const customToastsList = customToasts
+    // Avoid rendering transaction status toast.
+    // This is rendered individually in case of canceled/error transaction
+    .filter(
+      (customToast: CustomToastType) =>
```
StanislavSava, line 149: This logic is a bit contrived. We have custom toasts that can be opened and closed, and we use those for transaction toast statuses, but we don't render them together. Should we maybe add a separate slice/entry for customToast statuses, so we don't mix these up, since it's not the same logic?

src/components/TransactionStatusToast/TransactionStatusToast.tsx
StanislavSava: maybe FailedTransactionStatusToast?

src/components/TransactionStatusToast/TransactionStatusToast.tsx
```diff
@@ -0,0 +1,67 @@
...
+  message: string;
+  duration?: number;
+  onDelete?: () => void;
+};
+
+export const TransactionStatusToast: React.FC<ErrorToastProps> = ({
+  message,
+  duration = 30000,
+  onDelete,
+  type = 'warning'
```
StanislavSava, line 22: do we have an enum for these?

src/UI/SignTransactionsModals/components/TxStatusToast.tsx
CiprianDraghici: I'd keep the logic at high level. The same component is used, but conditioned by `signError` and `canceledTransactionsMessage`. Why should I use a single instance and conditionally pass the props?

src/UI/SignTransactionsModals/index.tsx
CiprianDraghici: You're right. I will refactor this part

src/UI/SignTransactionsModals/index.tsx
CiprianDraghici: simplified

src/UI/TransactionsToastList/index.tsx
CiprianDraghici: I tried to reuse useCustomHook as many as possible, but I think is better to create separate logic for this toast

src/components/TransactionStatusToast/StatusIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+import React from 'react';
+import { StatusIconType } from './types';
+
+type StatusIconProps = {
+  type: StatusIconType;
+};
+
+export const StatusIcon: React.FC<StatusIconProps> = ({ type }) => {
+  switch (type) {
+    case 'info':
```
grigorebudac, line 15: you can use the `StatusIconType` instead
also, maybe rename the `StatusIconType` to something that includes Enum so that it is more intuitive

grigorebudac (approved):

src/UI/SignTransactionsModals/components/TxStatusToast.tsx
```diff
@@ -0,0 +1,43 @@
...
+  onDelete
+}) => {
+  const message = useMemo(() => {
+    if (signError) {
+      return signError;
+    } else if (canceledTransactionsMessage) {
+      return canceledTransactionsMessage;
+    }
+
+    return 'Undefined transaction status';
```
StanislavSava, line 23: This should probably be something like "defaultErrorMessage"

src/UI/SignTransactionsModals/index.tsx
```diff
@@ -41,68 +22,39 @@ export const SignTransactionsModals = ({
...
+          className={className}
+        />
+      );
+    },
+    [verifyReceiverScam, className]
+  );
+
+  switch (loginMethod) {
+    case LoginMethodsEnum.ledger:
+      return renderScreen(ConfirmScreens.Ledger);
```
StanislavSava, line 50: instead of renderSCreen, wouldn't it be easier to just render the <ConfirmationScreen.../>

src/UI/SignTransactionsModals/components/TxStatusToast.tsx
```diff
@@ -0,0 +1,43 @@
...
+  canceledTransactionsMessage: string | null;
+  onDelete: () => void;
+};
+
+export const TxStatusToast: React.FC<TxStatusToastProps> = ({
+  signError,
+  canceledTransactionsMessage,
+  onDelete
+}) => {
+  const message = useMemo(() => {
```
StanislavSava, line 16: const message = signError || canceledTransactionsMessage || defaultMessage;

PR #308, opened 2022-07-07
2.0.0-beta.0

StanislavSava:


src/utils/logout.ts
```diff
@@ -48,14 +50,23 @@ export async function logout(
...
   }
 }
+
+function redirectToCallbackUrl(
+  callbackUrl?: string,
+  onRedirect?: (callbackUrl?: string) => void,
+  isWalletProvider?: boolean
+) {
+  if (callbackUrl && !isWalletProvider) {
+    if (typeof onRedirect === 'function') {
```
grigorebudac, line 66: why not check against null?

PR #312, opened 2022-07-08
[+] Added the wallet-connect v2 implementation using erdjs-wallet-con…

radumojic:
[+] Added the wallet-connect v2 implementation using erdjs-wallet-connect-provider to 2.1.0-alpha.0

- A projectId is required for the WalletConnect V2 functionality - get a free one from https://cloud.walletconnect.com. After you have the generated projectId, you can pass it via the walletConnectV2ProjectId customNetworkConfig key
- Both walletconnect providers are imported from the erdjs-wallet-connect-provider 2.1.0-alpha.0 package
- A new login method is added in the LoginMethodsEnum: walletconnectv2
- To use the WalletConnect 2 connection, set the isWalletConnectV2 param on the WalletConnectLoginButton button to true


src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,243 @@
...
+  PairingTypes
+} from '@elrondnetwork/erdjs-wallet-connect-provider';
+import { useUpdateEffect } from 'hooks/useUpdateEffect';
+import {
+  getAccountProvider,
+  setAccountProvider
+} from 'providers/accountProvider';
+import { loginAction } from 'reduxStore/commonActions';
+import { useDispatch, useSelector } from 'reduxStore/DappProviderContext';
+import { isLoggedInSelector } from 'reduxStore/selectors/loginInfoSelectors';
```
StanislavSava, line 13: This file looks very similar to useWalletConnectLogin for v1, can we extract the code that is used by both hooks into a general hooks and use that one, instead of copy pasting the logic?

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -63,45 +86,68 @@ export const WalletConnectLoginContainer = ({
...
-              }}
-            />
+            {qrCodeSvg ? (
+              <div
+                className={generatedClasses.qrCodeSvgContainer}
+                dangerouslySetInnerHTML={{
+                  __html: qrCodeSvg
+                }}
+                style={{
+                  width: '15rem',
```
StanislavSava, line 143: Let's move this to a class

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -110,7 +156,7 @@ export const WalletConnectLoginContainer = ({
                   id='accessWalletBtn'
                   data-testid='accessWalletBtn'
                   className={generatedClasses.mobileLoginButton}
-                  href={uriDeepLink || undefined}
+                  href={uriDeepLink || walletConnectDeepLinkV2 || undefined}
```
StanislavSava, line 159: I think the extra undefined can be removed with some types adjustments

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -127,10 +173,47 @@ export const WalletConnectLoginContainer = ({
             ) : (
               <p className={generatedClasses.leadText}>{lead}</p>
             )}
+            {isWalletConnectV2 && activePairings && activePairings?.length > 0 && (
+              <div className={generatedClasses.pairsContainer}>
```
StanislavSava, line 177: I think this JSX can be moved into a separate component and mounted here, to abstract away some code

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,243 @@
...
+      error,
+      loginFailed,
+      isLoading: isLoading && !loginFailed,
+      isLoggedIn: isLoggedIn && !loginFailed
+    },
+    { uriDeepLink, walletConnectUri: wcUri, wcPairings }
+  ];
+};
+
+export default useWalletConnectV2Login;
```
CiprianDraghici, line 243: No need to export default. Please use only named exports

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,243 @@
+import { useRef, useState } from 'react';
```
CiprianDraghici, line 1: I see that this hook is 80% the same with useWalletConnectLogin. Could you extract the common logic to avoid code duplication?

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,243 @@
...
+      const wcUriWithToken = `${walletConnectUri}&token=${token}`;
+
+      setWcUri(wcUriWithToken);
+      dispatch(setTokenLogin({ loginToken: token }));
+    }
+
+    try {
+      await providerRef.current?.login({ approval });
+    } catch {
+      await initiateLogin();
```
CiprianDraghici, line 225: usually, every error should be logged. Please log the error here before calling `initiateLogin()`

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,243 @@
...
+      if (token) {
+        dispatch(setTokenLogin({ loginToken: token }));
+      }
+
+      try {
+        await providerRef.current?.login({ approval });
+      } catch {
+        await initiateLogin();
+      }
+    } catch {
```
CiprianDraghici, line 192: Would be very nice to log the original error before setting a custom error. It will help a lot on debugging

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,243 @@
...
+      const { approval } = await providerRef.current?.connect({
+        topic: pairing.topic
+      });
+      if (token) {
+        dispatch(setTokenLogin({ loginToken: token }));
+      }
+
+      try {
+        await providerRef.current?.login({ approval });
+      } catch {
```
CiprianDraghici, line 189: log the error also here

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  async function connectExisting(pairing: PairingTypes.Struct | undefined) {
+    if (!walletConnectV2RelayAddress || !walletConnectV2ProjectId) {
+      setError(WalletConnectV2Error.invalidConfig);
+      return;
+    }
+    if (!pairing || !pairing.topic) {
+      return;
```
CiprianDraghici, line 176: I would set an error here just like at line 172 to figure out the reason why can not continue to connect. At least log a console.error to help on debugging.

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -43,12 +47,31 @@ export const WalletConnectLoginContainer = ({
...
+    logoutRoute,
+    callbackRoute,
+    token,
     onLoginRedirect
   });
   const [qrCodeSvg, setQrCodeSvg] = useState<string>('');
   const isMobileDevice =
     platform?.os?.family === 'iOS' || platform?.os?.family === 'Android';
+  const activePairings =
+    isWalletConnectV2 && wcPairings && wcPairings?.length > 0
```
grigorebudac, line 71: I think that `&& wcPairings` can be deleted since you're using optional chaining

radumojic: needed for typescript for now. Without it we get the `object is possibly undefined ` even though the code wouldn't actually run that case

src/hooks/login/useWalletConnectV2Login.ts
radumojic: We will keep the hooks separate for now since WC2 is still in a beta phase and more breaking changes might apprear. This way any changes won't affect WC1 that is already in a stable release.

src/hooks/login/useWalletConnectV2Login.ts
radumojic: We will keep the hooks separate for now since WC2 is still in a beta phase and more breaking changes might apprear. This way any changes won't affect WC1 that is already in a stable release.

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -127,10 +169,25 @@ export const WalletConnectLoginContainer = ({
             ) : (
               <p className={generatedClasses.leadText}>{lead}</p>
             )}
+            {isWalletConnectV2 &&
```
arhtudormorar, line 172: could be united in one variable

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,249 @@
...
+  sessionExpired = 'Unable to connect to existing session'
+}
+
+export interface WalletConnectV2LoginHookCustomStateType {
+  uriDeepLink: string;
+  walletConnectUri?: string;
+  wcPairings?: PairingTypes.Struct[];
+}
+
+export type WalletConnectV2LoginHookReturnType = [
```
arhtudormorar, line 50: I think types could be better exported from types folder

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,249 @@
...
+
+      dispatch(loginAction(loginActionData));
+      optionalRedirect(callbackRoute, onLoginRedirect);
+    } catch (err) {
+      setError(WalletConnectV2Error.invalidAddress);
+      console.error(err);
+    }
+  }
+
+  async function initiateLogin(loginProvider = true) {
```
arhtudormorar, line 140: there are many functions in this file. You could create a folder with index, types and other functions to break it into smaller pieces

radumojic: kept them all in the same file as the other providers to keep consistency

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -43,12 +48,31 @@ export const WalletConnectLoginContainer = ({
     logoutRoute,
     callbackRoute,
     token,
-    shouldLoginUser: true,
+    onLoginRedirect
+  });
+  const [
+    initLoginWithWalletConnectV2,
+    connectExisting,
```
StanislavSava, line 55: This break a bit the hook signature. If we return 3 items for all other hooks, the 4 items will break a common abstraction on top of them. Can we add this to the 3rd object, as an extra, specific config?

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -43,12 +48,31 @@ export const WalletConnectLoginContainer = ({
...
+    logoutRoute,
+    callbackRoute,
+    token,
     onLoginRedirect
   });
   const [qrCodeSvg, setQrCodeSvg] = useState<string>('');
   const isMobileDevice =
     platform?.os?.family === 'iOS' || platform?.os?.family === 'Android';
+  const activePairings =
+    isWalletConnectV2 && wcPairings && wcPairings?.length > 0
```
StanislavSava, line 72: you can skip the middle check, wcPairing?.length > 0 will always be false if wcPairing is null or undefined

src/UI/walletConnect/WalletConnectLoginContainer/index.tsx
```diff
@@ -43,12 +48,31 @@ export const WalletConnectLoginContainer = ({
...
+      ? wcPairings.filter((pairing) => !!pairing.active && pairing.peerMetadata)
```
StanislavSava, line 73: I think it's cleaner to use the Boolean(pairing.active) approach

arhtudormorar (approved):

src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -0,0 +1,249 @@
...
+} from 'reduxStore/slices';
+import { LoginHookGenericStateType } from 'types';
+import { LoginMethodsEnum } from 'types/enums';
+import { logout } from 'utils';
+import { optionalRedirect } from 'utils/internal';
+
+interface InitWalletConnectV2Type {
+  logoutRoute: string;
+  token?: string;
+  callbackRoute?: string;
```
StanislavSava, line 33: Do we still need this file? Seems like a lot of logic has been moved to the above hook

PR #336, opened 2022-07-19
Remove the shouldRenderDefaultCss flag and keep styles overriding at CSS level

CiprianDraghici:


src/UI/PageState/index.tsx
```diff
@@ -29,30 +28,25 @@ export const PageState = ({
...
+    wrapper: classNames(
+      styles.state,
+      globalStyles.mAuto,
+      globalStyles.p4,
+      globalStyles.textCenter
+    ),
+    iconContainer: classNames(
+      `${globalStyles.iconState} ${globalStyles.mxAuto}`,
+      {
+        [`${iconBgClass}`]: Boolean(iconBgClass)
```
arhtudormorar, line 43: I think wrapping in template string is not necessary

arhtudormorar (approved):

PR #341, opened 2022-07-25
[+] Add some utils functions to get transaction/transactions link

CiprianDraghici:


src/utils/transactions/getTransactionLink.ts
```diff
@@ -0,0 +1,5 @@
+import { NetworkType } from 'types';
+
+export function getTransactionLink(network: NetworkType, address: string) {
```
StanislavSava, line 3: Not sure we need to pass in the whole network object, we could just pass in the explorerAddress

src/hooks/transactions/useGetTransactionLink.ts
```diff
@@ -0,0 +1,15 @@
+import { getTransactionLink } from 'utils';
+import { useCallback } from 'react';
+import { useGetNetworkConfig } from '../useGetNetworkConfig';
+
+export const useGetTransactionLink = () => {
```
StanislavSava, line 5: I don't think we need a hook for this. You can just create a simple function that accepts a address and an explorer address which is optional, and if the explorer address is not passed, extract it with explorerAddressSelector(store.getState())

StanislavSava (changes requested):
No need to make this into a hook, this is a simple abstraction for an api call

PR #356, opened 2022-08-04
Adapt denominate ad fix tests

arhtudormorar:

PR #357, opened 2022-08-04
Adapt denominate ad fix tests (#356)

arhtudormorar:
* [*] Add devDependecies react and better-sqlite3

* [*] Make Denominate independent from redux store

* Merge with development


package.json
```diff
@@ -36,6 +36,7 @@
     "@types/qs": "6.9.7",
     "@typescript-eslint/eslint-plugin": "5.0.0",
     "@typescript-eslint/parser": "5.14.0",
+    "better-sqlite3": "7.6.2",
```
StanislavSava, line 39: ugh

src/UI/Denominate/index.tsx
```diff
@@ -91,8 +94,14 @@ const DenominateComponent = (props: DenominateType) => {
...
 };
 
+/**
+ * @param props.egldLabel  if not provided, will fallback on **EGLD**
+ */
 export const Denominate = (props: DenominateType) => {
-  const egldLabel = props.egldLabel || getEgldLabel();
+  const egldLabel =
+    props.egldLabel ||
+    fallbackNetworkConfigurations[EnvironmentsEnum.mainnet].egldLabel; // EGLD
```
StanislavSava, line 103: I wonder if we should extract this into a simple constant and use it inside configs, like we do with chainIds

src/UI/Denominate/tests/denominateComponent.test.tsx
```diff
@@ -22,7 +22,8 @@ describe('Denominate component when decimals = 2', () => {
       value: '9999979999800000000000000',
       showLastNonZeroDecimal: false,
       showLabel: true,
-      decimals: 2
+      decimals: 2,
+      egldLabel: 'EGLD'
```
StanislavSava, line 26: We should probably import this from constants

src/UI/Denominate/index.tsx
arhtudormorar: it's only this place we are using it like this. We can extract it to a constant if there will be one more place like this

package.json
arhtudormorar: yes to fix tests for now

src/UI/Denominate/tests/denominateComponent.test.tsx
arhtudormorar: it's OK since it's a test file

CiprianDraghici (approved):

StanislavSava (approved):

PR #359, opened 2022-08-04
Fix/do not store login info and avoid optional redirect in case of cancelled login

CiprianDraghici:


src/hooks/login/useWalletConnectLogin.ts
```diff
@@ -123,6 +123,11 @@ export const useWalletConnectLogin = ({
       }
 
       const address = await provider.getAddress();
+      if (!address) {
```
StanislavSava, line 126: The logic looks slightly different : here if there is no address there is a return, in the others there's an if for the address. Maybe we should extract that logic into a function that accepts the address, loginMethod, callbackRoute and onLoginRedirect and do all that for all providers?

CiprianDraghici: In the other cases there is a finally {} block that stops the loading indicator. Here is not the case and we can stop the execution early. I am checking if I can extract some common logic

StanislavSava (approved):

PR #364, opened 2022-08-12
Feature/check transactions status using WebSocket and display the status in toasts

CiprianDraghici:


radumojic (approved):

package.json
```diff
@@ -138,6 +138,8 @@
     "react-redux": "8.0.2",
     "redux-persist": "6.0.0",
     "reselect": "4.0.0",
-    "shx": "0.3.4"
+    "shx": "0.3.4",
+    "socket.io": "4.5.1",
```
arhtudormorar, line 142: check remove

src/components/TransactionsTracker/useTransactionsTracker.ts
```diff
@@ -1,40 +1,40 @@
...
 import { getTransactionsByHashes as defaultGetTxByHash } from 'apiCalls/transactions';
 import { GetTransactionsByHashesType } from 'types';
-import { useCheckTransactionStatus, useGetPendingTransactions } from 'hooks';
-import { getIsTransactionPending } from 'utils';
+import {
+  useCheckTransactionStatus,
+  useGetAccountInfo,
+  useGetNetworkConfig,
+  useRegisterWebsocketListener
+} from 'hooks';
```
arhtudormorar, line 9: from hooks/index because in tests jest appends .js extension and might not find hooks

src/components/TransactionsTracker/useTransactionsTracker.ts
```diff
@@ -1,40 +1,40 @@
...
 export interface TransactionsTrackerType {
   getTransactionsByHash?: GetTransactionsByHashesType;
 }
 
 export function useTransactionsTracker(props?: TransactionsTrackerType) {
-  const { pendingTransactionsArray } = useGetPendingTransactions();
+  const { network } = useGetNetworkConfig();
+  const { address } = useGetAccountInfo();
+  const checkTransactionStatus = useCheckTransactionStatus();
+  const checkTransactionStatusRef = useRef(checkTransactionStatus);
```
StanislavSava, line 19: why do we need this in a ref?

src/components/TransactionsTracker/useTransactionsTracker.ts
```diff
@@ -1,40 +1,40 @@
...
-        checkTransactionStatus({ getTransactionsByHash });
-      }, 2000);
-    } else {
-      clearInterval(intervalRef.current);
-    }
-    return () => {
-      clearInterval(intervalRef.current);
-    };
-  }, [pendingBatches]);
+    checkTransactionStatusRef.current = checkTransactionStatus;
```
StanislavSava, line 32: why the dance with ref and useEffect?

src/components/TransactionsTracker/useTransactionsTracker.ts
```diff
@@ -1,40 +1,40 @@
...
+  }, [checkTransactionStatus]);
+
+  useRegisterWebsocketListener({
```
StanislavSava, line 35: There's a tricky situation here: this is a hook, and if you call this hook multiple times or maybe in different components, this will run multiple times, opening a lot of connections. 

We need to rethink this a bit: either run this hook only once per mounting the app, or, even better, inside the useRegisterWebsocketListener, make a list of listeners with keys to them and don't register a listener if it already has one

src/components/TransactionsTracker/useClosureRef.ts
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from 'react';
+
+export const useClosureRef = <T>(value: T) => {
+  const refVal = useRef<T>(value);
+
+  useEffect(() => {
+    refVal.current = value;
+  }, [value]);
+
+  return refVal;
```
StanislavSava, line 10: probably should return refVal.current, to avoid reading .current everywhere

src/components/TransactionsTracker/useTransactionsTracker.ts
```diff
@@ -1,40 +1,35 @@
...
-
-  const pendingBatches = pendingTransactionsArray.filter(
-    ([sessionId, transactionBatch]) => {
-      const isPending =
-        sessionId != null && getIsTransactionPending(transactionBatch.status);
-      return isPending;
-    }
-  );
+  const onMessage = () => {
+    checkTransactionStatusCallbackRef.current({
```
StanislavSava, line 25: Don't think we need the "ref" suffix

src/hooks/websocketListener/useGetWebsocketUrl.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from 'react';
+import { getWebsocketUrl } from 'utils';
+
+export function useGetWebsocketUrl(apiAddress: string) {
```
StanislavSava, line 4: let's use the same way we do API calls in dapp-core, check getAccountFromApi, for example

src/hooks/websocketListener/useRegisterWebsocketListener.ts
```diff
@@ -0,0 +1,72 @@
+import { useCallback, useEffect } from 'react';
+import { io } from 'socket.io-client';
+import { websocketConnection } from './websoketConnection';
```
StanislavSava, line 3: typo in file name

src/hooks/websocketListener/useRegisterWebsocketListener.ts
```diff
@@ -0,0 +1,72 @@
...
+  address
+}: UseRegisterWebsocketListenerPropsType) {
+  const { network } = useGetNetworkConfig();
+
+  const initializeWebsocketConnection = useCallback(
+    retryMultipleTimes(
+      async () => {
+        // If there are many components that use this hook, the initialize method is triggered many times.
+        // To avoid multiple connections to the same endpoint, we have to guard the initialization before the logic started
+        websocketConnection.status = 'pending';
```
StanislavSava, line 27: this should probably be in an enum

src/hooks/websocketListener/useRegisterWebsocketListener.ts
```diff
@@ -0,0 +1,72 @@
...
+        websocketConnection.current = io(websocketUrl, {
+          forceNew: true,
+          reconnectionAttempts: RECONNECTION_ATTEMPTS,
+          timeout: TIMEOUT,
+          query: {
+            address
+          }
+        });
+
+        websocketConnection.status = 'completed';
```
StanislavSava, line 45: probably should be in an enum too

src/hooks/websocketListener/useRegisterWebsocketListener.ts
```diff
@@ -0,0 +1,72 @@
...
+        delay: RETRY_INTERVAL
+      }
+    ),
+    [address, onMessage]
+  );
+
+  useEffect(() => {
+    if (
+      address &&
+      websocketConnection.status === 'not_initialized' &&
```
StanislavSava, line 60: same here

src/components/TransactionsTracker/useClosureRef.ts
CiprianDraghici: It would not work. I already tried this solution

src/hooks/websocketListener/useGetWebsocketUrl.ts
CiprianDraghici: I already added a "utils" function called: getWebsocketUrl. We have a pair (hook-utils function for getting websocket url)

StanislavSava (approved):

PR #387, opened 2022-08-25
Feature/add fallback mechanism to check transactions status in case of down ws

CiprianDraghici:


src/components/TransactionsTracker/useTransactionsTracker.ts
```diff
@@ -30,4 +33,11 @@ export function useTransactionsTracker(props?: TransactionsTrackerType) {
     onMessage,
     address
   });
+
+  useEffect(() => {
+    timeoutRef.current = setTimeout(() => onMessage(), 30000);
```
StanislavSava, line 38: I wonder if this shouldn't be an interval, if transaction takes longer than 30 seconds, so we can do multiple calls

src/components/TransactionsTracker/useTransactionsTracker.ts
```diff
@@ -12,6 +13,8 @@ export interface TransactionsTrackerType {
 }
 
 export function useTransactionsTracker(props?: TransactionsTrackerType) {
+  const timeoutRef = useRef<NodeJS.Timeout>();
```
StanislavSava, line 16: I don't think we need a ref for this, you can keep the reference in a local variable inside useEffect

src/components/TransactionsTracker/useTransactionsTracker.ts
StanislavSava, line 38: You can just do setTimeout(onMessage)

StanislavSava (approved):

PR #397, opened 2022-08-26
Fix/ledger sign transactions UI and api endpoints

CiprianDraghici:


src/UI/SignTransactionsModals/SignWithDeviceModal/SignStep.tsx
```diff
@@ -87,8 +87,10 @@ export const SignStep = ({
     isLastTransaction && !waitingForDevice ? 'Sign & Submit' : signBtnLabel;
   signBtnLabel = continueWithoutSigning ? 'Continue' : signBtnLabel;
 
+  const nftId = `${tokenId}-${nonce}`;
+
   const { tokenDecimals, tokenAvatar } = useGetTokenDetails({
-    tokenId: currentTransaction.transactionTokenInfo.tokenId
+    tokenId: nonce ? nftId : tokenId
```
StanislavSava, line 93: this will do a false positive if nonce is 0, is this what we want? or do we need to do a nonce != null?

StanislavSava: Also I think we should add a little commentary here, it's a bit confusing what this does

CiprianDraghici: the nonce is a string

CiprianDraghici: done

StanislavSava (approved):

arhtudormorar (approved):

PR #426, opened 2022-09-21
[fix-safari-send-transaction] Add delay before redirection so we don't get canceled 

mihairadulescu:
This is a fix for a bug we got on inspire.art for safari.
Seems that safari browser cancels all xhr request made when writing in window.location ..... like on navigation cancel all xhr request..
The nasty thing is that all axios calls will trigger a network error, simple as that, no reasons why or just know it was aborted.

This is the easiest fix I could apply.


arhtudormorar:
Has been resolved in v.2.1.0-rc3

PR #465, opened 2022-11-04
Fix window and document undefined (#464)

arhtudormorar:
* Add document and window guard for Next.js


razvantomegea (approved):

ilyich-erd:
This PR is not actually fixing the issue with SSR. You should add checks over `window`. Something like 
```js
typeof window !== "undefined"
```
should do the thing.

PR #477, opened 2022-11-16
[*] Progress bar fix and interface renaming

MiroMargineanu:


src/UI/Progress/index.tsx
```diff
@@ -9,7 +9,7 @@ import styles from './progressStyles.scss';
 
 const TOAST_PROGRESS_KEY = 'toastProgress';
 
-export interface ProgressProps extends WithClassnameType {
+export interface ProgressPropsType extends WithClassnameType {
```
arhtudormorar, line 12: breaking change. better write TODO: rename in v3

MiroMargineanu: Reverted and noted. 👍

arhtudormorar (approved):

Issue #358, opened 2022-08-04
Building error when using dapp-core in a nextjs app


joaquin-alfaro:
Thanks for the library of components, it is great and really helpful when making elrond dapps.  
 
I have to make a SSR dapp and I am planning to use [nextjs](nextjs.org), but the building of the app fails when using any 
 component of **dapp-core**

# Expected behavior
The building of a [nextjs](nextjs.org) app using components from dapp-core finish successfully and generates the folder .next with the required files to start the app.

# Actual behavior
The building of the [nextjs](nextjs.org) app fails with the following error:
```
> Build error occurred
/nextjs-with-elrond-dapp-core/node_modules/@elrondnetwork/dapp-core/UI/index.js:1
"use strict";import{a as n}from"../__chunks__/chunk-Z2GXZLB7.js";import{a as l}from"../__chunks__/chunk-2R3H532S.js";
             ^^^^^^

SyntaxError: Cannot use import statement outside a module
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1033:15)
    at Module._compile (node:internal/modules/cjs/loader:1069:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Module._load (node:internal/modules/cjs/loader:827:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at 272 (/nextjs-with-elrond-dapp-core/.next/server/pages/elrond-page.js:22:28)
    at __webpack_require__ (/nextjs-with-elrond-dapp-core/.next/server/webpack-runtime.js:25:42) {
  type: 'SyntaxError'
}
```

# Steps to reproduce the behavior
1. Create nextjs app
```sh
npx create-next-app@latest --typescript
npm run build
```
2. Add dapp-core dependencies
```sh
npm install @types/react@18.0.14 @types/react-dom@18.0.5 --save-dev
npm install @elrondnetwork/dapp-core @elrondnetwork/erdjs @elrondnetwork/erdjs-network-providers
```
3. Use a dapp-core component in a page
```javascript
import type { NextPage } from 'next'
import * as DappUI from '@elrondnetwork/dapp-core/UI'
  
const Elrond: NextPage = () => {
    const { WebWalletLoginButton } = DappUI
    return (
        <div>
[...]

PR #505, opened 2022-12-13
:bug: nextjs support

maximechbt:
This PR allows dapp-core to be used in nextJS, except for UI components.

It's the usage of SCSS imports in TS/JS file that generates many document.createStyle in the build files which is not compatible in ssr.
So, if it's okay for you, we should avoid using SCSS import outside of UI components to allow NextJs users to use dapp-core, at least for hooks and services.

This is a quick fix to show the problem, but there is surely an abstraction issue, a service should not be aware of the display and therefore not return ui classes.

There may also be something to do at build time to make this system SSR compatible...but I don't know much about that.

### DIFF
--- a/src/services/transactions/signTransactions.ts
+++ b/src/services/transactions/signTransactions.ts
@@ -1,5 +1,4 @@
 import BigNumber from 'bignumber.js';
-import globalStyles from 'assets/sass/main.scss';
 import { GAS_LIMIT } from 'constants/index';
 
 import { accountBalanceSelector } from 'reduxStore/selectors/accountInfoSelectors';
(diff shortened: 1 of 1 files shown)


michavie:
The abstraction issue you're calling out is certainly of concern (also see #358), which is why many fall back to using erdjs directly for their dapps and build a wrapper around the DAppProviders.

It would be nice for DX for sure to have clean, headless, non-heavily opinionated libraries as a layer on top, and then incompatibilities with NextJs or any other consuming framework should be a problem of the past

CiprianDraghici (approved):

CiprianDraghici:
Please check this dApp template
https://github.com/multiversx/mx-template-nextjs-dapp

PR #508, opened 2022-12-14
[+] Linkify

razvantomegea:
- substituted anchorme with linkify.js due to security concerns;
- optimized getScamFlag.


src/types/suspiciousInfo.ts
```diff
@@ -0,0 +1,15 @@
+import { ScamInfoType } from './account.types';
+
+export interface SuspiciousInfo {
```
arhtudormorar, line 3: suspicous is too vague. Maybe replace with SuspiciousLinkType
then also rename file to suspiciousLink.ts
and below SuspiciousLinkPropsType

src/utils/transactions/transactionInfoHelpers/tests/getScamFlag.test.ts
```diff
@@ -1,11 +1,11 @@
 import { getScamFlag } from '../getScamFlag';
 
 describe('scamDetect tests', () => {
-  const output = '[Message hidden due to suspicious content - ';
+  const output = 'Message hidden due to suspicious content - ';
   const strings: { [key: string]: string[] } = {
-    '👉 link.com': ['👉 link.com', ''],
+    '👉 link.com': ['👉 http://link.com', ''],
```
arhtudormorar, line 6: tests should not change because scams do not necessarily put http inside their messages

razvantomegea: The http is for the output, not the input.

arhtudormorar (approved):

PR #515, opened 2022-12-19
Stop login on nativeAuth block failed

arhtudormorar:
- extend tests


src/UI/extension/ExtensionLoginButton/tests/helpers/mockExtensionProvider.ts
```diff
@@ -2,14 +2,39 @@ import { testAddress } from '__mocks__';
 
 export const ExtensionProvider = {
```
CiprianDraghici, line 3: I think this is a mock. Maybe it can be moved in a `__mocks__ `folder

src/UI/extension/ExtensionLoginButton/tests/helpers/mockExtensionProvider.ts
```diff
@@ -2,14 +2,39 @@ import { testAddress } from '__mocks__';
...
+
     const provider = {
       init: async () => true,
-      login: async () => true,
+      login: async ({ token }: { token?: string }) => {
+        if (token) {
+          signature =
+            'e4c98dd01020118b13db9dd5db9e5b56ff0c4a0141306918a9d3eea964a21ada5d566f58cdf6c921ed3405bf5685d1e87545dbcc86ea3c27a43aa3abee8c2b0e';
+        }
+        return {
```
CiprianDraghici, line 14: The login method is async, then you should return a promise.
Eg. `Promise.resolve({...})`

src/hooks/login/useExtensionLogin.ts
```diff
@@ -59,6 +59,12 @@ export const useExtensionLogin = ({
 
       if (hasNativeAuth && !token) {
         token = await loginService.getNativeAuthLoginToken();
+
+        // fetchig block failed
+        if (!token) {
+          console.warn('Login cancelled.');
```
CiprianDraghici, line 65: What about a more suggestive text here, something like: "Session expired" ?

src/hooks/login/useLedgerLogin.ts
```diff
@@ -129,6 +129,11 @@ export function useLedgerLogin({
 
     if (hasNativeAuth && !token) {
       token = await loginService.getNativeAuthLoginToken();
+      // fetchig block failed
+      if (!token) {
+        console.warn('Login cancelled.');
```
CiprianDraghici, line 134: Please see my comment above. I see that the log message is used for all login methods. If you decide to change the text, please take into consideration also the other places

src/hooks/login/useLoginService.ts
```diff
@@ -103,6 +108,9 @@ export const useLoginService = (config?: OnProviderLoginType['nativeAuth']) => {
   }) => {
     const loginToken = await getNativeAuthLoginToken();
     tokenRef.current = loginToken;
+    if (!loginToken) {
+      return;
```
CiprianDraghici, line 112: Would be better to log a warning or an error message here

arhtudormorar: this function is not ready. It must be changed by @StanislavSava or @andreigiura

CiprianDraghici (approved):

PR #559, opened 2023-01-11
partially support for nextjs

CiprianDraghici:


src/hooks/login/useWalletConnectLogin.ts
```diff
@@ -79,6 +79,10 @@ export const useWalletConnectLogin = ({
   }, [provider]);
 
   useUpdateEffect(() => {
+    if(!tokenToSign) {
```
arhtudormorar, line 82: token is not always present, so this is not a valid condition

ioandanpopa (approved):

src/reduxStore/store.ts
```diff
@@ -1,5 +1,6 @@
 import { configureStore } from '@reduxjs/toolkit';
-import { createSubscription } from 'react-redux/es/utils/Subscription';
+// @ts-ignore
+import { createSubscription } from 'react-redux/lib/utils/Subscription';
```
arhtudormorar, line 3: can you import the type? if not possible, its' OK like this

PR #740, opened 2023-04-24
[FIX] Fixed web wallet login when token is not provided

razvantomegea:


src/hooks/login/useWebWalletLogin.ts
```diff
@@ -53,7 +52,7 @@ export const useWebWalletLogin = ({
         expires: expires
       };
 
-      if (hasNativeAuth && !token) {
```
arhtudormorar, line 56: this was correct, it should not fetch a native auth token if hasNativeAuth flag is not set

PR #774, opened 2023-05-10
Fix cancel transactions flow with web wallet provider

CiprianDraghici:
### Issue/Feature
Cancel transactions action throw an error in dApp
The cancel transaction toast is not displayed
### Reproduce
Issue exists on version `2.13.1` of sdk-dapp.

### Root cause
Web wallet dApp does not return the `transaction` param through URL in case of cancel which causes the client app to crash because of that (`sdk-dapp` execute URL sanitization and it expects the `transaction` param which is not provided)
### Fix
Avoid app crashing on cancel flow
Display cancel transaction toast
### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/hooks/transactions/useParseSignedTransactions.tsx
+++ b/src/hooks/transactions/useParseSignedTransactions.tsx
@@ -49,7 +49,8 @@ export function useParseSignedTransactions(
             removeParams: [
               ...Object.keys(SignedMessageQueryParamsEnum),
               WALLET_PROVIDER_CALLBACK_PARAM,
-              WALLET_SIGN_SESSION
+              WALLET_SIGN_SESSION,
+              'address'
             ],
             search
           });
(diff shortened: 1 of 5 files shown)


src/hooks/transactions/useParseSignedTransactions.tsx
```diff
@@ -35,11 +42,7 @@ export function useParseSignedTransactions(
             })
           );
           onAbort();
-          const [transaction] = signedTransactions;
-          removeTransactionParamsFromUrl({
-            transaction,
-            search
-          });
+          dispatch(setSignTransactionsCancelMessage(TRANSACTION_CANCELLED));
```
arhtudormorar, line 45: + clear url data like signSession & searchData.status

razvantomegea (changes requested):

arhtudormorar (approved):

PR #775, opened 2023-05-10
Fix handle unsent batch transactions

CiprianDraghici:
### Issue/Feature
Some batches are not processed by the service and have missing transaction hashes
Duplicate batch transactions send
### Reproduce
Issue exists on version `2.13` of sdk-dapp.

### Root cause
Batch transactions service
Error in the batch transactions send flow (duplicate send)
### Fix
Handle this case in order to resolve the toast with the correct status (notExecuted)
### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/hooks/transactions/batch/tracker/useUpdateBatch.ts
+++ b/src/hooks/transactions/batch/tracker/useUpdateBatch.ts
@@ -1,14 +1,17 @@
-import { refreshAccount } from 'utils/account/refreshAccount';
-import { useGetBatches } from '../useGetBatches';
-import { sequentialToFlatArray } from 'utils/transactions/batch/sequentialToFlatArray';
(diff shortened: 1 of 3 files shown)


src/hooks/transactions/batch/tracker/useUpdateBatch.ts
```diff
@@ -81,6 +86,28 @@ export function useUpdateBatch() {
             })
           );
         }
+      } else {
```
arhtudormorar, line 89: there seem to be many nested statements here. You could consider some small function to handle this logic

razvantomegea (approved):

arhtudormorar (approved):

PR #788, opened 2023-05-16
Fix nonce management

CiprianDraghici:
### Issue
Incorrect nonce management on signing batch transactions with web wallet
### Reproduce
Issue exists on version `2.` of sdk-dapp.

### Root cause
WebWallet dApp doesn't know what the client dApp will do with the signed transactions (will send the signed transactions or not). Then, when a set of transactions arrives in the web wallet to be signed,  the web wallet takes the current account nonce from the store and computes the transactions nonces related to this. If you send the next batch (the previous one has not been executed yet) the nonce from the account is the same, then the next batch will consider the same nonce (account nonce) and will provide the transactions signed with a wrong nonce to the client dApp.
### Additional changes
- remove unused `preventNonceUpdate`
### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/hooks/transactions/helpers/useSetTransactionNonces.ts
+++ b/src/hooks/transactions/helpers/useSetTransactionNonces.ts
@@ -12,11 +12,13 @@ const setTransactionNonces = (
     return transactions;
   }
 
-  const firstTxNonce = transactions[0].getNonce().valueOf();
-  const computedNonce = firstTxNonce > latestNonce ? firstTxNonce : latestNonce;
-
   return transactions.map((tx: Transaction, index: number) => {
-    tx.setNonce(computedNonce + index);
+    const nextNonce = latestNonce + index;
+
+    const txNonce = tx.getNonce().valueOf();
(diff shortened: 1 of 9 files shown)


src/hooks/transactions/helpers/useSetTransactionNonces.ts
```diff
@@ -8,8 +8,17 @@ const setTransactionNonces = (
...
+  if (transactions.length === 0) {
+    return transactions;
+  }
+
   return transactions.map((tx: Transaction, index: number) => {
-    tx.setNonce(latestNonce + index);
+    const nextNonce = latestNonce + index;
+
+    const txNonce = tx.getNonce().valueOf();
+    const computedNonce = txNonce > nextNonce ? txNonce : nextNonce;
```
arhtudormorar, line 19: add comment
`// stop replacing nonce if transaction is configured with a higher nonce than the existing one`

arhtudormorar (approved):

PR #803, opened 2023-05-26
Added the sign step indexing

MiroMargineanu:
### Issue/Feature
There's no way currently to distinguish which transaction you're signing when signing multiple transactions inside the Web Wallet.

### Reproduce
Issue exists on version `2.14.1` of sdk-dapp.

### Root cause
Markup was not added to support this scenario.

### Fix
Added an indexing system in the title of the step that counts the step you're on and the total steps.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests


yarn.lock
```diff
@@ -13595,8 +13595,8 @@
     "shallowequal" "^1.1.0"
     "throttle-debounce" "^3.0.1"
 
-"react@file:../mx-wallet-dapp/node_modules/react":
-  "resolved" "file:../mx-wallet-dapp/node_modules/react"
+"react@file:../mx-exchange-dapp/node_modules/react":
```
arhtudormorar, line 13598: pls delete and regenereate yarn lock

MiroMargineanu: Regenerated. 👍

CiprianDraghici (approved):

PR #859, opened 2023-07-05
Fix cancel transaction toast. Display the toast in the same container as the transaction toasts.

CiprianDraghici:
### Issue
The cancel transaction toast is displayed over the pending transactions toasts

### Reproduce
Issue exists on version `2.17.2` of sdk-dapp.

### Root cause
The cancel transaction toast is rendered in a separate container from transactions toasts.
### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests


src/UI/SignTransactionsModals/components/TransactionStatusToast.tsx
```diff
@@ -37,12 +39,17 @@ export const TransactionStatusToast = ({
...
-      type={type}
-      onDelete={onDelete}
-      className={className}
-    />
-  );
+  useEffect(() => {
+    addNewCustomToast({
+      toastId: `${CANCEL_TRANSACTION_TOAST_ID}-${Date.now()}`,
+      title: 'Transaction canceled',
+      duration: 20000,
```
arhtudormorar, line 46: this should be configurable

arhtudormorar (approved):

PR #873, opened 2023-07-24
[FIX] Login redirect timeout

razvantomegea:
### Issue
- `safeRedirect` method was never called because of incorrect condition;
- removed `setTimeout` around another `setTimeout` in `optionalRedirect`;
- removed unnecessary `setTimeout` around `onLoginRedirect`, because the timout was set to 0.

### Additional changes
- Jest error when rendering component with svg (fixed `jest.config`)
- Refactoring of `LedgerLoginContainer`

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[x] Unit tests


src/UI/ledger/LedgerLoginButton/tests/LedgerLoginButton.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+        <LedgerLoginButton
+          wrapContentInsideModal={false}
+          callbackRoute={CALLBACK_ROUTE}
+          nativeAuth
+        />
+      )
+    });
+
+    const setTokenLoginSpy = jest.spyOn(actions, 'setTokenLogin');
+    jest.spyOn(Date, 'now').mockReturnValue(1690184313013);
```
arhtudormorar, line 104: 1690184313013 --> better obtain this from a human readable date

src/UI/ledger/LedgerLoginContainer/LedgerLoading/LedgerLoading.tsx
```diff
@@ -33,6 +33,7 @@ export const LedgerLoading = ({
   return (
     <div
       className={classNames(styles.ledgerLoadingWrapper, ledgerLoadingWrapper)}
+      data-testid='ledgerLoading'
```
arhtudormorar, line 36: I think we should start using dataTestIds from enums

src/hooks/login/useLoginService.ts
```diff
@@ -51,8 +51,7 @@ export const useLoginService = (config?: OnProviderLoginType['nativeAuth']) => {
 
   const getNativeAuthLoginToken = async () => {
```
arhtudormorar, line 52: is async still needed here?

src/utils/internal/optionalRedirect.ts
```diff
@@ -17,19 +17,17 @@ export function optionalRedirect({
 
   const hasOnLoginRedirect = typeof onLoginRedirect === 'function';
 
-  const timeout = hasOnLoginRedirect ? 0 : DEFAULT_TIMEOUT;
-
   if (shouldRedirect && callbackRoute != null) {
-    setTimeout(() => {
```
arhtudormorar, line 23: this is no longer pushing the execution in the event loop and might have side effects in some scenarios especially when dealing with browser window redirect which is a different thread from the normal JS execution thread

PR #919, opened 2023-09-13
clear extension bgr when cancel tx from dapp

andreigiura:
### Reproduce
Issue exists on version `2.20.2` of sdk-dapp.
DeFi wallet would not clear the transaction when cancelled from DApps

### Root cause
cancelAction was removed when canceling transactions from DApps
### Fix
Add back cancelAction when cancelling a transaction and remove it from transaction toast discard

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/UI/SignTransactionsModals/components/ConfirmationScreen.tsx
+++ b/src/UI/SignTransactionsModals/components/ConfirmationScreen.tsx
@@ -52,8 +52,7 @@ export const ConfirmationScreen = ({
 
   const transactionStatusToastProps: TransactionStatusToastType = {
     signError,
-    canceledTransactionsMessage,
-    onDelete: handleClose
+    canceledTransactionsMessage
   };
 
   const shouldShowTransactionStatusToast =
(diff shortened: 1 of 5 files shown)


src/UI/SignTransactionsModals/components/ConfirmationScreen.tsx
```diff
@@ -52,8 +52,7 @@ export const ConfirmationScreen = ({
 
   const transactionStatusToastProps: TransactionStatusToastType = {
     signError,
-    canceledTransactionsMessage,
-    onDelete: handleClose
```
CiprianDraghici, line 56: onDelete: handleClose

not needed anymore ?

CiprianDraghici (approved):

src/UI/SignTransactionsModals/components/ConfirmationScreen.tsx
andreigiura: yes, needed for the fix. onDelete is optional and is kept and passed by other components, but not in this case

PR #927, opened 2023-09-16
Add network option when refreshing a native auth token

andreigiura:
### Reproduce
-
### Root cause
-
### Fix
-

### Additional changes
Add  network option when refreshing a native auth token.  Allows of refreshing a native auth token, with a new network just before network change.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/services/nativeAuth/helpers/getLatestBlockHash.ts
+++ b/src/services/nativeAuth/helpers/getLatestBlockHash.ts
@@ -49,7 +49,8 @@ const getLatestBlockHashFromServer = retryMultipleTimes(
 export async function getLatestBlockHash(
   apiUrl: string,
   blockHashShard?: number,
-  getBlockHash?: () => Promise<string>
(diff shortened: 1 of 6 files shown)


CiprianDraghici (approved):

src/services/nativeAuth/helpers/getLatestBlockHash.ts
```diff
@@ -59,7 +60,8 @@ export async function getLatestBlockHash(
   if (
     cachedResponse.current != null &&
     currentTimestampMs <
-      cachedResponse.current.timestamp * 1000 + cachingDurationMs
+      cachedResponse.current.timestamp * 1000 + cachingDurationMs &&
```
MiroMargineanu, line 63: It might be easier to read if all these conditions were split up in variables.

MiroMargineanu (approved):

PR #943, opened 2023-10-03
[FIX] Fixed `ModalContainer` not working in `StrictMode`

razvantomegea:
### Issue
Modals do not appear in the Ledger and WalletConnect login buttons when using `StrictMode`

### Reproduce
Add `StrictMode` around the main `App` component and try to login using `LedgerLoginButton` or `WalletConnectButton`

### Root cause
The `useDappModal` manages the state in redux store with 2 effect hooks that alter the visibility of the modal.

### Fix
Remove modal management from redux and handle the configs locally.

### Additional changes
Redirect with page refresh after login with `window.location.assign`, instead of just `history.pushState`. The latter was not working well.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[x] Unit tests


src/UI/DappModal/dappModal.types.ts
```diff
@@ -15,6 +15,5 @@ export interface DappModalConfig {
 }
 
 export interface DappModalInteractionConfig {
-  openOnMount?: boolean;
```
arhtudormorar, line 18: Is this not breaking?

src/UI/DappModal/hooks/useDappModal.ts
```diff
@@ -1,71 +1,29 @@
...
 
-export interface UseDappModalProps {
-  config?: DappModalInteractionConfig;
-  modalConfig?: DappModalConfig;
-}
+import { DappModalInteractionConfig } from '../dappModal.types';
 
-export const useDappModal = (props?: UseDappModalProps) => {
-  const dispatch = useDispatch();
-  const visible = useIsDappModalVisible();
```
arhtudormorar, line 21: I think this was persisting the visibility in some cases

src/utils/logout.ts
```diff
@@ -89,7 +89,7 @@ function redirectToCallbackUrl(
     if (typeof onRedirect === 'function') {
       onRedirect(callbackUrl);
     } else {
-      safeRedirect({ url: callbackUrl, shouldForcePageReload: true });
```
arhtudormorar, line 92: why this change?

src/utils/redirect.ts
```diff
@@ -19,13 +17,6 @@ export const safeRedirect = ({
         return;
       }
 
-      // Navigate to callbackUrl without page refresh if we are in the same origin
```
arhtudormorar, line 22: It would have been better if this was in a separate PR

src/UI/DappModal/hooks/useDappModal.ts
razvantomegea: It is not necessary to have one centralised modal managed with redux

src/UI/DappModal/dappModal.types.ts
razvantomegea: No, because we have the `visible` props/flag which was already doing that

src/utils/logout.ts
razvantomegea: Will make another PR for this

src/utils/redirect.ts
razvantomegea: Will make another PR for this

src/utils/logout.ts
razvantomegea: https://github.com/multiversx/mx-sdk-dapp/pull/944

src/utils/redirect.ts
razvantomegea: https://github.com/multiversx/mx-sdk-dapp/pull/944

src/UI/DappModal/hooks/useDappModal.ts
CiprianDraghici: I think the idea was to have this type of modal visible once at a time

razvantomegea: This implies the component is also rendered once, like `<SignTransactionsModals />`, not multiple times in the app

CiprianDraghici (approved):

PR #944, opened 2023-10-03
[FIX] Added page reload on login redirect

razvantomegea:
### Issue
After login, the page does not redirect correctly to the specified `callbackRoute`

### Reproduce
Issue exists on version `2.22.1` of sdk-dapp.

### Root cause
Redirect is done with `window.history.pushState` instead of `window.location.assign` which seems to have undesirable side effects.

### Fix
Replace `window.history.pushState` with `window.location.assign` in the `safeRedirect` helper

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[x] Unit tests

### DIFF
--- a/src/utils/logout.ts
+++ b/src/utils/logout.ts
@@ -89,7 +89,7 @@ function redirectToCallbackUrl(
     if (typeof onRedirect === 'function') {
       onRedirect(callbackUrl);
     } else {
-      safeRedirect({ url: callbackUrl, shouldForcePageReload: true });
+      safeRedirect({ url: callbackUrl });
     }
   }
 }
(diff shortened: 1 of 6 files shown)


arhtudormorar (approved):

src/utils/logout.ts
```diff
@@ -89,7 +89,7 @@ function redirectToCallbackUrl(
     if (typeof onRedirect === 'function') {
       onRedirect(callbackUrl);
     } else {
-      safeRedirect({ url: callbackUrl, shouldForcePageReload: true });
+      safeRedirect({ url: callbackUrl });
```
CiprianDraghici, line 92: are you sure you have to remove the property `shouldForcePageReload` ?

CiprianDraghici (approved):

PR #973, opened 2023-11-28
Fixed logout redirect loop

razvantomegea:
### Issue
When performing a refresh after token expiration or when the Ledger is inactive (`provider` is no empty), there is an infinite loop of redirects to the `unlock` route

### Reproduce
Issue exists on version `2.24.0` of sdk-dapp.

### Root cause
When we are logout without explicitly doing so (idle time, token expiration, Ledger disconnect, etc.), we ar still logged in the store, but the provider is of type `none`. 

In this scenario, there was a condition the forced always a redirect to the `unlock` route, without actually doing a logout action in the `redux` store.

### Fix
Do a logout action regardless of being logged in or not. Skip only the `provider.logout` action if the provider is not initialised, `null`, or of type `none`.

### Additional changes
Make the `shouldAttemptRelogin` property from the `logout` hook default to `false`, instead of `true`, because relogin attempt is a special case, not a general one.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests


arhtudormorar (approved):

src/utils/logout.ts
```diff
@@ -28,46 +27,39 @@ const broadcastLogoutAcrossTabs = (address: string) => {
 export async function logout(
   callbackUrl?: string,
   onRedirect?: (callbackUrl?: string) => void,
-  shouldAttemptRelogin = true
+  shouldAttemptRelogin = false
```
CiprianDraghici, line 30: `Boolean(getWebviewToken()) `instead of false

CiprianDraghici (approved):

PR #1013, opened 2024-01-08
Prevent logout before webview login

razvantomegea:
### Issue

The web wallet logouts and reloads the page when trying to login via webview.

### Reproduce
Issue exists on version `2.26.6` of sdk-dapp.

### Root cause
The logout action is triggered from sdk-dapp and the `shouldAttemptRelogin` parameter is set to `false`, because the `webviewToken` is missing. This is caused by the logout actions dispatch triggered before the `nativeAuth` login action.

### Fix
Prevent sending the logout action before initiating the login action within the webview login.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/services/nativeAuth/helpers/loginWithNativeAuthToken.ts
+++ b/src/services/nativeAuth/helpers/loginWithNativeAuthToken.ts
@@ -4,18 +4,20 @@ import { store } from 'reduxStore/store';
 import { LoginMethodsEnum } from 'types';
 import { decodeNativeAuthToken } from './decodeNativeAuthToken';
 
-export function loginWithNativeAuthToken(token: string, dispatch?: any) {
+export function loginWithNativeAuthToken(
+  token: string,
+  dispatch = store.dispatch
+) {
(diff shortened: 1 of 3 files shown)


src/services/nativeAuth/helpers/loginWithNativeAuthToken.ts
```diff
@@ -1,30 +1,28 @@
-import { loginAction, logoutAction } from 'reduxStore/commonActions';
+import { loginAction } from 'reduxStore/commonActions';
 import { setTokenLogin } from 'reduxStore/slices';
 import { store } from 'reduxStore/store';
 import { LoginMethodsEnum } from 'types';
 import { decodeNativeAuthToken } from './decodeNativeAuthToken';
 
 export function loginWithNativeAuthToken(token: string, dispatch?: any) {
```
arhtudormorar, line 7: dispatch = store.dispatch

arhtudormorar (approved):

PR #1051, opened 2024-02-16
Add getHasNativeAuth

mgavrila:
### Issue
On local development, if nativeAuth is not set, it will still use nativeAuth to sign token for login when using WalletConnect.

### Reproduce
Issue exists on version `2.28.7-alpha.3` of sdk-dapp.

### Root cause
origin property from nativeAuth which is being set on development mode from wallet makes walletConnect to always use the nativeAuth feature.

### Fix
Add getHasNativeAuth function which will clear the origin property if it is used in development mode.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/hooks/login/useWalletConnectV2Login.ts
+++ b/src/hooks/login/useWalletConnectV2Login.ts
@@ -16,7 +16,7 @@ import {
 } from 'reduxStore/selectors/networkConfigSelectors';
 import { setWalletConnectLogin } from 'reduxStore/slices';
 import { LoginMethodsEnum } from 'types/enums.types';
-import { checkNativeAuth } from 'utils/checkNativeAuth';
+import { getHasNativeAuth } from 'utils/getHasNativeAuth';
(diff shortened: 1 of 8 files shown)


src/hooks/login/useWalletConnectV2Login.ts
```diff
@@ -74,7 +75,7 @@ export const useWalletConnectV2Login = ({
   customRequestMethods = []
 }: InitWalletConnectV2Type): WalletConnectV2LoginHookReturnType => {
   const dispatch = useDispatch();
-  const hasNativeAuth = nativeAuth != null;
+  const hasNativeAuth = checkNativeAuth(nativeAuth);
```
arhtudormorar, line 78: rename to `getHasNativeAuth()`

arhtudormorar (approved):

PR #1070, opened 2024-03-12
Fix experimental webview provider initialization

CiprianDraghici:
### Issue
The experimental webview provider doesn't notify the parent when is loaded.
### Reproduce
Issue exists on version `2.29.0-beta.0` of sdk-dapp.

### Root cause
Wrong implementation of the init() function

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/providers/accountProvider.ts
+++ b/src/providers/accountProvider.ts
@@ -28,7 +28,6 @@ export function setAccountProvider<TProvider extends ProvidersType>(
 
 export function setExternalProvider(provider: IDappProvider) {
   externalProvider = provider;
-  externalProvider.init?.();
 }
 
 export function setExternalProviderAsAccountProvider() {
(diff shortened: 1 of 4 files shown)


src/providers/accountProvider.ts
```diff
@@ -28,6 +28,7 @@ export function setAccountProvider<TProvider extends ProvidersType>(
 
 export function setExternalProvider(provider: IDappProvider) {
   externalProvider = provider;
+  externalProvider.init?.();
```
arhtudormorar, line 31: I would call this specifically in experimental provider

arhtudormorar (approved):

razvantomegea (approved):

andreigiura (approved):

PR #1104, opened 2024-03-26
Fixed cross-window ledger re-login when device is locked

razvantomegea:
### Issue
The wallet automatically logs out (with redirect to different host) and replies with cancelled when logged in with a locked or unavailable ledger device.

### Reproduce
Issue exists on version `2.29.0-beta.18` of sdk-dapp.

### Root cause
When ledger is busy and we are logged in, the providerType appears as wallet and ledger initialisation fails. This causes the logout to be triggered and a redirect to the wallet URL, without trying to re-login.

### Fix
Prevent wallet redirects within the wallet provider `logout` method if the provider is not initialised

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/utils/logout.ts
+++ b/src/utils/logout.ts
@@ -81,17 +81,8 @@ export async function logout(
     matchPath(location.pathname, callbackPathname) ||
     (isWalletProvider && isProviderInitialised)
   ) {
-    console.log('Prevent redirects');
     preventRedirects();
   }
-
-  console.log('Logout', {
-    address,
-    providerType,
-    callbackPathname,
-    isProviderInitialised
-  });
-  debugger;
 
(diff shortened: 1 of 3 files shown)


src/utils/logout.ts
```diff
@@ -81,9 +81,18 @@ export async function logout(
...
     preventRedirects();
   }
 
+  console.log('Logout', {
+    address,
+    providerType,
+    callbackPathname,
+    isProviderInitialised
+  });
+  debugger;
```
arhtudormorar: 🪵

arhtudormorar (approved):

PR #1209, opened 2024-07-19
Fix unexpected logout in webview (xPortal Hub and Web Wallet Hub)

CiprianDraghici:
### Issue
dApps are automatically closed due to an unexpected logout triggered by a slower provider login/initialization in combination with search params manipulation or eventual redirects
### Reproduce
Issue exists on version `2.34.2` of sdk-dapp.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests


src/providers/experimentalWebViewProvider/ExperimentalWebviewProvider.ts
```diff
@@ -40,9 +39,8 @@ export class ExperimentalWebviewProvider implements IDappProvider {
   };
 
   logout = async () => {
-    const response = await this._provider.logout();
-    await logoutFromDApp();
-    return response;
+    store.dispatch(logoutAction());
```
arhtudormorar, line 42: why not like before?

CiprianDraghici: because it would generate a loop
dApp calls explicitly logout function -> logout function calls provider.logout (this line here) -> provider logout function calls logout function who calls again the provider.logout...

arhtudormorar (approved):

PR #1217, opened 2024-08-06
Prevent setting polling interval when websocket is present

razvantomegea:
### Issue

Infinite loop of get account requests are seen in the network tab

### Root cause
`refreshAccount()` is called at every 1.5s (`ROUND_DURATION / 2`) within the `useTransactionsTracker.ts`.

### Fix
Do not set pollingInterval if we have websocket initialised.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests


src/hooks/transactions/useTransactionsTracker.ts
```diff
@@ -21,11 +25,22 @@ export function useTransactionsTracker(props?: TransactionsTrackerType) {
 
   useRegisterWebsocketListener(onMessage);
 
-  useEffect(() => {
+  const setPollingInterval = async () => {
+    const websocketUrl = await getWebsocketUrl(network.apiAddress);
+
+    if (websocketUrl != null) {
```
arhtudormorar, line 31: I think the condition here should be:
- do not set interval if subscription to WS event was successful

(ws addres might be down so backup mechanism should enter)

arhtudormorar (approved):

PR #1221, opened 2024-08-06
Fixed error when calling `fungibleFromBigInteger` with invalid input

razvantomegea:
### Issue
There is an error thrown that: `Invalid argument amountAsBigInteger`

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/utils/account/refreshAccount.tsx
+++ b/src/utils/account/refreshAccount.tsx
@@ -8,14 +8,18 @@ import { getLatestNonce } from './getLatestNonce';
 const setNewAccount = async () => {
   try {
     const address = await getAddress();
+
     try {
       const account = await getAccount(address);
+
       if (account != null) {
         const accountData = {
           ...account,
           nonce: getLatestNonce(account)
         };
+
         store.dispatch(setAccount(accountData));
+
         return accountData;
       }
     } catch (e) {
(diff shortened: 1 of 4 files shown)


src/utils/account/refreshAccount.tsx
```diff
@@ -24,30 +28,36 @@ const setNewAccount = async () => {
...
 
-      const initialized = await provider.init();
-      if (!initialized) {
-        return;
-      }
-      return setNewAccount();
-    } catch (e) {
-      console.error('Failed initializing provider ', e);
+  try {
+    if (!provider.init) {
```
arhtudormorar, line 47: web wallet does not have init but it's valid

src/utils/operations/formatAmount.ts
```diff
@@ -38,11 +38,23 @@ export function formatAmount({
   return (
     pipe(modInput as string)
       // format
-      .then(() =>
-        TokenTransfer.fungibleFromBigInteger('', modInput as string, decimals)
-          .amountAsBigInteger.shiftedBy(-decimals)
-          .toFixed(decimals)
-      )
+      .then(() => {
+        try {
```
arhtudormorar, line 42: no, we should not do this here

PR #1234, opened 2024-08-12
Fixed transaction toasts are not cleared

razvantomegea:
### Issue
When having multiple pending transactions, the toasts are not cleared properly/on time. This causes the web wallet/dapp to work very slow.

### Root cause
The timer that is supposed to delete the not pending toast is always reset on re-render. The re-render happens when multiple transactions are in progress or if other transactions are initiated while others are in progress.

### Fix

- Prevent clearing the timer on re-render/unmount;
- Add a ref and assign the timer to that ref;
- Add a timer only if the transaction is no longer pending and there is no existing timer;
- Fix the logic of setting the timer to remove toasts.

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests


CiprianDraghici (approved):

src/UI/TransactionsToastList/components/TransactionToast/hooks/useTransactionToast.ts
```diff
@@ -78,20 +77,13 @@ export const useTransactionToast = ({
...
+    if (isPending || !lifetimeAfterSuccess || timeoutRef.current) {
       return;
     }
 
-    const timeout = setTimeout(() => {
+    timeoutRef.current = setTimeout(() => {
       handleDeleteToast();
     }, lifetimeAfterSuccess);
-
-    return () => {
```
CiprianDraghici, line 92: why is removed ?

CiprianDraghici (changes requested):
Please check the timeout removal

CiprianDraghici (approved):

PR #1272, opened 2024-09-24
Fixed multisig login

razvantomegea:
### Issue
Dapp is logged in with the account address instead of account multisig address

### Root cause
The account is refreshed and the address is used and replaces the multisig address.

### Fix
Ignore account refresh when logged in with multisig

### Contains breaking changes
[x] No

[] Yes

### Updated CHANGELOG
[x] Yes

### Testing
[x] User testing
[] Unit tests

### DIFF
--- a/src/components/ProviderInitializer/helpers/processModifiedAccount.ts
+++ b/src/components/ProviderInitializer/helpers/processModifiedAccount.ts
@@ -46,7 +46,5 @@ export const processModifiedAccount = async <
     loginService.setTokenLoginInfo({ signature, address });
   }
 
-  const account = await getAccount(accountAddress);
-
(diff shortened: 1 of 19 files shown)


src/components/ProviderInitializer/helpers/processModifiedAccount.ts
```diff
@@ -46,7 +46,5 @@ export const processModifiedAccount = async <
     loginService.setTokenLoginInfo({ signature, address });
   }
 
-  const account = await getAccount(accountAddress);
```
arhtudormorar, line 49: export const getAccount = (address?: string) => getAccountFromApi(address);

was correct before

PR #1304, opened 2024-11-11
Ability to show transaction toast on demand

razvantomegea:


src/UI/TransactionsToastList/components/CustomToast/CustomToast.tsx
```diff
@@ -11,6 +13,19 @@ export const CustomToast = (props: CustomToastPropsType) => {
     return <CustomComponentToast {...props} />;
   }
 
+  if (props.transaction) {
+    const transaction = props.transaction as any;
```
CiprianDraghici, line 17: I'd use `SignedTransactionType` instead of any

razvantomegea: I tried, but it throws errors. The `TransactionToast` and `CustomToast` accespt both `SignedTransactionType` and `ServerTransactionType` which have different implementation of `transaction`. It complicates things very much to change the types and not worth it. Using simply `any` here does the job without too much hustle

CiprianDraghici: `SignedTransactionType | ServerTransactionType`. If you want to cast to a specific type you can use `as unknown as ServerTransactionType`.
A specific type is desired because you know what to expect from that variable when you read the code first time. 'any' should be used just in exceptional situations because it breaks the code safeness

CiprianDraghici (approved):

src/hooks/transactions/useSignTransactionsWithDevice.tsx
```diff
@@ -160,7 +160,8 @@ export function useSignTransactionsWithDevice(
     if (!transaction) {
       return null;
     }
-    return await connectedProvider.signTransaction(transaction);
+
+    return connectedProvider.signTransaction(transaction);
```
arhtudormorar, line 164: Mi se părea mai clar cu await (chiar dacă codul merge la fel)

src/utils/account/signMessage.ts
```diff
@@ -38,7 +38,7 @@ export const signMessage = async ({
     );
   }
 
-  return await provider.signMessage(signableMessage, {
+  return provider.signMessage(signableMessage, {
```
arhtudormorar, line 41: Același comentariu

arhtudormorar (approved):

PR #1311, opened 2024-11-20
Added transaction data decoder

razvantomegea:


package.json
```diff
@@ -184,6 +184,7 @@
     "qs": "6.10.3",
     "react-idle-timer": "5.0.0",
     "react-redux": "8.0.2",
+    "react-select": "5.8.3",
```
arhtudormorar, line 187: why is it needed?

src/UI/TransactionData/components/TransactionDataDecode.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    onDecode(displayValue);
+
+    if (validationWarnings.length > 0) {
+      onDecodeError(validationWarnings);
+    }
+  }, [method, data]);
+
+  return (
+    <Select
```
arhtudormorar, line 62: why not use simple HTML select?

PR #1312, opened 2024-11-25
Fixed BigNumber.js scoping

MiroMargineanu:
### Issue
The `BigNumber` configuration was not properly cleaned up in the `stringIsFloat` helper function, thus overwriting the configuration in other helper functions too, creating a race condition.

### Reproduce
Issue exists on version `3.0.12` of sdk-dapp.

### Root cause
The cause was the `BigNumber.set` method called globally in the `stringIsFloat` helper function.

### Fix
Create a clone of the `BigNumber` instance and scope it locally, and only change the settings at a local function level.

### Contains breaking changes
- [x] No
- [ ] Yes

### Updated CHANGELOG

- [ ] No
- [x] Yes

### Testing

- [x] User testing
- [ ] Unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


arhtudormorar (commented):
changelog

arhtudormorar (approved):

PR #1319, opened 2024-11-26
Updated get network config from API logic

razvantomegea:
The network config is fetched from `/dapp/config` regardless. We need to fetch the config only if the `chainId` is missing or if we need the `roundDuration`. The `roundDuration` is used in (sovereign) chains that do not have websocket and rely on regular polling intervals to refresh the data.

### DIFF
--- a/src/components/ProviderInitializer/ProviderInitializer.tsx
+++ b/src/components/ProviderInitializer/ProviderInitializer.tsx
@@ -60,6 +60,11 @@ import {
   getIframeProvider
 } from './helpers';
 import { useSetLedgerProvider } from './hooks';
+import {
+  DEVNET_CHAIN_ID,
+  MAINNET_CHAIN_ID,
+  TESTNET_CHAIN_ID
+} from '../../constants';
 
 let initalizingLedger = false;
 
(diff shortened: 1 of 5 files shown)


src/components/ProviderInitializer/ProviderInitializer.tsx
```diff
@@ -120,6 +125,16 @@ export function ProviderInitializer() {
   }, [ledgerAccount, isLoggedIn, ledgerData]);
 
   async function refreshNetworkConfig() {
+    const shouldGetConfig =
+      !network.chainId ||
+      ![DEVNET_CHAIN_ID, TESTNET_CHAIN_ID, MAINNET_CHAIN_ID].includes(
```
CiprianDraghici, line 130: please place a comment here why this is needed

arhtudormorar: this could also be resolved with suggestive variable naming (at least 50% of the explenation)

arhtudormorar:
maybe add PR comments

arhtudormorar (approved):

PR #1321, opened 2024-11-26
Fixed Ledger login

razvantomegea:
### Issue
Ledger login with USB no longer works. Instant error occurs and only after multiple clicks it works

### Root cause
The ledger account is no longer cleared when error occurs because the provider init failed. The accounts are fetched without an intialised provider, so, it fails instantly. Many race conditions occur.

### Fix
Wait for the Ledger/HW provider to be initialized on component mount and fetch the accounts otherwise.
Continue the regular login after this and show errors only if the loading state is true (loading state means the user started to login)

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -30,7 +30,7 @@ jobs:
         run: yarn install
 
       - name: Run tests
-        run: yarn test
+        run: yarn test --silent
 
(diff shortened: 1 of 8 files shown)


CiprianDraghici (approved):

arhtudormorar (approved):

arhtudormorar:
maybe add PR description

PR #1338, opened 2024-12-05
Ledger Address Table Pagination Feature

MiroMargineanu:
### Feature
Added proper pagination to the Ledger provider address table, that allows navigating to custom pages of the table instead of the usual back and forth which was very slow. We've capped this to 500 pages allowing up to 5,000 addresses to be used. This can be easily modified.

### Reproduce
Authenticate on any third party application leveraging on the package and try navigating between pages, it'll be very difficult since the only options are back and forth.

### Additional Information
Feature doesn't exist on version `3.0.23` of `sdk-dapp`.
This feature contains a slight breaking change because it now requires the `onGoToSpecificPage` property callback to be sent to the `AddressTable` component in order to handle navigation to pages other than the siblings.

### Contains breaking changes

- [ ] No
- [x] Yes

### Updated CHANGELOG

- [ ] No
- [x] Yes

### Testing

- [x] User testing
- [x] Unit tests


src/UI/Pagination/helpers/getPagination/getPagination.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface GetPaginationType {
+  currentPage: number;
+  totalPages: number;
+}
+
+export const getPagination = ({
+  currentPage,
+  totalPages
+}: GetPaginationType) => {
+  const maxSlots = 7;
```
arhtudormorar, line 15: why not above function?

src/hooks/login/useLedgerLogin.ts
```diff
@@ -341,7 +342,7 @@ export const useLedgerLogin = ({
         return onLoginFailed(failInitializeErrorText);
       }
 
-      if (accounts.length === 0 || startIndex > 0) {
+      if (accounts.length === 0 || startIndex >= 0) {
```
arhtudormorar, line 345: why the change?

MiroMargineanu: Because when navigating to the second page of the table and then going back to the first, it'd no longer fetch the first page of addresses, and instead it'd show the ones from the second page. This can be replicated even now, on the live version.

src/UI/Pagination/helpers/getPagination/getPagination.ts
MiroMargineanu: Updated. 👍

arhtudormorar (approved):

PR #1367, opened 2025-01-31
Added Optional Receiver Trimming Option

MiroMargineanu:
### Feature
Add the possibility to optionally not trim the receiver inside the `ConfirmReceiver` components. This is useful for the situations where you don't want to show a trimmed address in case of sending to a scam address by mistake.

### Reproduce
Feature doesn't exist on version `3.2.0` of `sdk-dapp`.

### Contains breaking changes
- [x] No
- [ ] Yes

### Updated CHANGELOG
- [ ] No
- [x] Yes

### Testing
- [x] User testing
- [ ] Unit tests

### DIFF
--- a/src/UI/SignTransactionsModals/SignWithDeviceModal/components/components/ConfirmReceiver/ConfirmReceiver.tsx
+++ b/src/UI/SignTransactionsModals/SignWithDeviceModal/components/components/ConfirmReceiver/ConfirmReceiver.tsx
@@ -120,20 +120,16 @@ const ConfirmReceiverComponent = ({
             </span>
           )}
 
-          {shouldTrimReceiver && (
-            <CopyButton
-              text={receiver}
-              className={styles?.receiverCopy}
-              copyIcon={customCopyIcon}
-            />
-          )}
+          <CopyButton
+            text={receiver}
+            className={styles?.receiverCopy}
+            copyIcon={customCopyIcon}
+          />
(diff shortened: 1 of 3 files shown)


src/UI/SignTransactionsModals/SignWithDeviceModal/components/components/ConfirmReceiver/ConfirmReceiver.tsx
```diff
@@ -110,16 +120,20 @@ const ConfirmReceiverComponent = ({
             </span>
           )}
 
-          <CopyButton
-            text={receiver}
-            className={styles?.receiverCopy}
-            copyIcon={customCopyIcon}
-          />
+          {shouldTrimReceiver && (
```
razvantomegea, line 123: Why show copybutton only when receiver should be trimmed? What does that have to do with this?

razvantomegea (changes requested):

src/UI/SignTransactionsModals/SignWithDeviceModal/components/components/ConfirmReceiver/ConfirmReceiver.tsx
MiroMargineanu: Restored. 👍

razvantomegea (approved):

PR #1378, opened 2025-03-10
Fixed websocket connection and falback mechanism

razvantomegea:
### Issue
Websocket not reconnecting after logout and the fallback mechanism (polling interval) is not triggered.

### Reproduce

1. Login with one address;
2. Logout
3. Login back with another address.

Actual Result: "Websocket reconnecting..." message appears (it should not when logged out)
Expected Result: "Websocket connected" message appears (after logged in again)

### Root cause
- The old address is checked within the functions (hooks closure);
- There is not check for previous websocket status to see the change differences in the fallback mechanism;
- Websocket connection status is not reactive to trigger fallback mechanism on webscoket connection failure.

### Fix
- Fetch the address directly from store in the functions (`setTimeout` and `setInterval`);
- Make the `WebsocketConnection` reactive.

### Additional changes
- Added ` useWebsocketPollingFallback` to reuse the polling interval fallback mechanism

### Contains breaking changes

- [x] No
- [ ] Yes

### Updated CHANGELOG

- [ ] No
- [x] Yes

### Testing

- [x] User testing
- [ ] Unit tests

### DIFF
--- a/src/hooks/transactions/useTransactionsTracker/useWebsocketPollingFallback.ts
+++ b/src/hooks/transactions/useTransactionsTracker/useWebsocketPollingFallback.ts
@@ -87,7 +87,5 @@ export const useWebsocketPollingFallback = ({
     };
   }, [address, websocketStatus, onPoll, usedPollingInterval]);
 
-  return {
-    clearFallbackTimer
-  };
+  return clearFallbackTimer;
 };


src/hooks/transactions/useTransactionsTracker/useWebsocketPollingFallback.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      onPoll();
+    }, usedPollingInterval);
+
+    return () => {
+      clearFallbackTimer();
+    };
+  }, [address, websocketStatus, onPoll, usedPollingInterval]);
+
+  return {
```
arhtudormorar, line 90: could return just a value but can stay

arhtudormorar (approved):

PR #1383, opened 2025-03-11
Fixed cancel action is not called on all providers

razvantomegea:
### Issue
Multiple extension windows/popups appear after cancelling a transaction from the "Confirm on Defi Wallet" modal. Also, no sign transaction modal appears in the extension.

### Root cause
Cancel action is not called when "Close" button is click in the "Confirm on Defi Wallet" modal.

### Contains breaking changes

- [x] No
- [ ] Yes

### Updated CHANGELOG

- [ ] No
- [x] Yes

### Testing

- [x] User testing
- [ ] Unit tests


radumojic (approved):

src/hooks/transactions/useCancelActionAndDispose.ts
```diff
@@ -0,0 +1,35 @@
...
+      return;
+    }
+
+    try {
+      const typedProvider = provider as unknown as {
+        cancelAction: () => Promise<boolean> | boolean;
+        dispose: () => Promise<void> | void;
+      };
+
+      await typedProvider.cancelAction();
```
arhtudormorar, line 26: does it crash on ledger?
what abouw web wallet URL ?

PR #1578, opened 2025-09-16
Added extrasApi field

mgavrila:
### Issue
- extrasApi field missing from network

### Reproduce

Issue exists on version `5.1.5` of sdk-dapp.

### Contains breaking changes

- [x] No
- [ ] Yes

### Updated CHANGELOG

- [ ] No
- [x] Yes

### Testing

- [x] User testing
- [ ] Unit tests

### DIFF
--- a/src/constants/network.constants.ts
+++ b/src/constants/network.constants.ts
@@ -18,6 +18,7 @@ export const fallbackNetworkConfigurations: Record<
     xAliasAddress: 'https://devnet.xalias.com',
     apiAddress: 'https://devnet-api.multiversx.com',
     explorerAddress: 'http://devnet-explorer.multiversx.com',
+    extrasApi: 'https://devnet-extras-api.multiversx.com',
     apiTimeout: '4000',
     roundDuration: 6000
   },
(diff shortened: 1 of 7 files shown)


src/providers/ProviderFactory.ts
```diff
@@ -125,11 +125,7 @@ export class ProviderFactory {
     setAccountProvider(dappProvider);
 
     const shouldClearInitiatedLogins = (
-      [
-        ProviderTypeEnum.crossWindow,
-        ProviderTypeEnum.metamask,
-        ProviderTypeEnum.passkey
-      ] as readonly ProviderType[]
+      [ProviderTypeEnum.crossWindow] as readonly ProviderType[]
     ).includes(type);
```
MiroMargineanu, line 129: Does this still need to be an array and check against the type with `.includes()` since there's just one item now?

MiroMargineanu (approved):

src/constants/network.constants.ts
```diff
@@ -18,6 +18,7 @@ export const fallbackNetworkConfigurations: Record<
     xAliasAddress: 'https://devnet.xalias.com',
     apiAddress: 'https://devnet-api.multiversx.com',
     explorerAddress: 'http://devnet-explorer.multiversx.com',
+    extrasApi: 'https://devnet-extras-api.multiversx.com',
```
arhtudormorar: add it as optional

MiroMargineanu (approved):

arhtudormorar (approved):

PR #1596, opened 2025-09-27
Enhance logout handling in useNativeAuthLogout and improve token expiration checks

razvantomegea:
### Issue/Feature

Fixed race condition causing immediate logout when opening a new tab and logging in quickly. Users were being automatically logged out after 3-5 seconds in new tabs, but the same behavior worked correctly when staying in the same tab (even after clearing cache/refreshing).

### Reproduce

1. Open a new browser tab
2. Navigate to the dapp and login immediately
3. User gets automatically logged out after 3-5 seconds
4. This issue does NOT occur when:
   - Staying in the same tab and logging in
   - Clearing browser cache and refreshing in the same tab
   - Waiting before logging in

### Root cause

The race condition occurs due to how React components and browser tabs work independently:

**New Tab Behavior:**

- Creates a completely new React component instance
- All `useRef` values are initialized fresh
- Timeout IDs start from scratch in the new JavaScript context
- Each tab maintains its own isolated state and refs

**Same Tab Behavior:**

- Reuses the same browser context but reinitializes the React application
- Component unmounting and remounting happens in the same JavaScript execution environment
- Timeout IDs may have continuity or overlap

The specific issue was in the `useNativeAuthLogout` hook where:

1. New tab opens → Fresh component instance created
2. User logs in immediately → `useEffect` triggers with fresh token data
3. `plannedLogoutRef.current` is empty (fresh ref) → condition passes
4. Timeout is set for token expiration
5. Race condition occurs between the newly set logout timeout and existing token expiration check

### Fix

Enhanced the `useNativeAuthLogout` hook with proper cleanup and validation:

1. **Added proper cleanup**: Clear timeouts and reset planned logout ref when conditions change
2. **Added minimum threshold**: Require at least 1 second remaining before setting logout timeout
3. **Improved timeout management**: Clear any existing timeout before setting new one
4. **Added safety check**: Double-check planned logout key before executing logout
5. **Updated dependencies**: Added `secondsUntilExpires` to dependency array

Key changes:

- Clear timeouts when `address` or `isWebviewLogin` conditions change
- Minimum 1-second threshold to prevent immediate logout for barely expired tokens
- Double-check conditions before logout to prevent stale timeout execution
- Better cleanup of refs and timeouts

### Additional changes
Revert commit 86d63047 with removal of `login-session-expiry`

### Contains breaking changes

- [x] No
- [ ] Yes

### Updated CHANGELOG

- [ ] No
- [x] Yes

### Testing

- [x] User testing
- [ ] Unit tests

### DIFF
--- a/src/reduxStore/middlewares/loginSessionMiddleware.ts
+++ b/src/reduxStore/middlewares/loginSessionMiddleware.ts
@@ -1,0 +1,59 @@
+import { PayloadAction } from '@reduxjs/toolkit';
+import throttle from 'lodash.throttle';
+import { LOGOUT_ACTION_NAME } from 'constants/index';
+import { invalidateLoginSession } from 'reduxStore/slices';
+import { RootState } from 'reduxStore/store';
+import { getNewLoginExpiresTimestamp, setLoginExpiresAt } from 'storage/local';
+import { storage } from 'utils/storage';
+import { localStorageKeys } from 'utils/storage/local';
+
+const whitelistedActions = [LOGOUT_ACTION_NAME];
+let invalidatedSession = false;
+
+const throttledSetNewExpires = throttle(() => {
+  setLoginExpiresAt(getNewLoginExpiresTimestamp());
+}, 5000);
+
+export const loginSessionMiddleware =
+  (store: any) =>
+  (next: (action: PayloadAction) => void) =>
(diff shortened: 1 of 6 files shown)


arhtudormorar (commented):
revert to prev ?

mgavrila (approved):

PR #1604, opened 2025-10-02
Fixed explorer links not working in transactions tables

iuliacimpeanu:
### Issue
Receiver shard link and transaction value link were not directing to explorer.

### Contains breaking changes

- [x] No
- [ ] Yes

### Updated CHANGELOG

- [ ] No
- [x] Yes

### Testing

- [x] User testing
- [ ] Unit tests


src/controllers/TransactionsTableController/TransactionsTableController.ts
```diff
@@ -39,6 +39,14 @@ export class TransactionsTableController {
       })
     );
 
+    const getExplorerUrl = (relativePath?: string) => {
```
arhtudormorar, line 42: it's strange to have an optional argument that makes the function return undefined if not provided

PR #1607, opened 2025-10-06
Added ability to bypass sender validation for specific origins in `checkIsValidSender`

razvantomegea:


arhtudormorar (approved):

src/hooks/transactions/helpers/checkIsValidSender.ts
```diff
@@ -10,6 +11,11 @@ export const checkIsValidSender = (
     return true;
   }
 
+  // Allow sender to be different than the logged in address if the origin is multiversx.com
+  if (safeWindow?.location?.origin?.endsWith('wallet.multiversx.com')) {
```
MiroMargineanu, line 15: Is it correct to have Wallet logic inside the package which should be agnostic?

razvantomegea: Yes. It is an internal requirement

